PR #20, opened 2021-09-02
Add HDF5 support for InfiniteCanonicalMPS

hershsingh:


mtfishman:
Over all looks good, thanks! A few things:

1. You need to add `HDF5` as a dependency. In the Julia REPL, type:
```julia
julia> using Pkg

julia> Pkg.activate("ITensorInfiniteMPS")

julia> Pkg.add("HDF5")
```
After doing this, you should see `HDF5` added to the file `Project.toml`. Then, add `using HDF5` to the `src/ITensorInfiniteMPS.jl` file here: https://github.com/mtfishman/ITensorInfiniteMPS.jl/blob/ee89e088d03300d1bf60b838172dbb2b4231e283/src/ITensorInfiniteMPS.jl#L3-L14
2. You can see that the "format-check" Github Action failed. I use the package `JuliaFormatter` to enforce a standard format for the code. First install the package with `using Pkg; Pkg.add("JuliaFormatter")`. Then in the base directory of the package `~/.julia/dev/ITensorInfiniteMPS`, run `using JuliaFormatter; format(".")` in the REPL, which will automatically format the code.
3. Please add a test for reading and writing.

test/test_readwrite.jl
```diff
@@ -0,0 +1,34 @@
+using ITensors
+using ITensorInfiniteMPS
+using HDF5
```
mtfishman, line 3: Try changing this to:
```julia
using ITensorInfiniteMPS.HDF5
```
This will use the "internal" version of HDF5 that will already be available in `ITensorInfiniteMPS`, so it doesn't need to try to separately install `HDF5`.

`using HDF5` can work, but you would have to add `HDF5` as a test dependency by activating the environment in `~/.julia/dev/ITensorInfiniteMPS/test` and doing `Pkg.add("HDF5")`.

hershsingh: Thanks! Just did that and pushed a commit.

PR #28, opened 2021-12-31
Update installation instructions

hershsingh:
Pkg.add(..) takes the package url as a "url" keyword argument now


mtfishman:
Thanks!

PR #30, opened 2022-01-12
Add `eigsolve_tol` as a keyword argument for the eigensolver

hershsingh:
This exposes the Krylov eigensolver tolerance `krylov_tol` as a keyword argument of the `vumps` function.

### DIFF
--- a/.github/workflows/format_check.yml
+++ b/.github/workflows/format_check.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        julia-version: [1.3.0]
+        julia-version: [1.7.1]
         julia-arch: [x86]
         os: [ubuntu-latest]
     steps:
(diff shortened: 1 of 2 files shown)


mtfishman:
Looks good, thanks! I'll wait for tests to pass and then merge.

hershsingh:
Looks like it failed the format test. Let me run JuliaFormatter on this.

hershsingh:
I ran JuliaFormatter on the entire directory. Not sure why GitHub's check is failing on `src/infinitecanonicalmps.jl`

mtfishman:
That's strange. I think I've seen that before, and the problem was that the Github CI workflow was using a different version of JuliaFormatter than I was using locally. I see that it is set to use Julia 1.3.0 here:
https://github.com/ITensor/ITensorInfiniteMPS.jl/blob/5b765f1a7a46939251e91f8c5233d0c7e1f4ad02/.github/workflows/format_check.yml#L16

You could try increasing that to Julia 1.7.1 and see if that helps. If that doesn't work than it's not worth spending time on, we can fix it later.

hershsingh:
Ok, yes changing the version works. Thanks!

PR #31, opened 2022-01-23
Fix VUMPS for multi-site unit cells and nonlocal interactions

LHerviou:
As discussed, fixes the creation of InfiniteMPS for more than two sites in the unit cell.


src/infinitemps.jl
```diff
@@ -69,8 +69,10 @@ function Base.getindex(l::InfiniteITensorSum, n1n2::Tuple{Int,Int})
   @assert n2 == n1 + 1
   return l.data[n1]
 end
-nsites(h::InfiniteITensorSum) = length(l.data)
-
+nsites(h::InfiniteITensorSum) = length(h.data)
+#Gives the range of the Hamiltonian. Useful for better optimized contraction in VUMPS
+nranges(h::InfiniteITensorSum) = order.(h.data)÷2
+nrange(h::InfiniteITensorSum, n::Int64) = order(h.data[n])÷2
```
mtfishman, line 75: I guess this points to `nsites` being an overloaded term, since `nsites` would also be a good name for this. In fact, I don't I like `nrange` for this since the there could be a two-site interaction with a long range. Maybe `nsite_support` would be a better name for this.

mtfishman: Though I may be misinterpreting what this function is supposed to do, in a general case.

LHerviou: Fair point. I implicitly assumed that local hamiltonians are always involves contiguous sets of sites, it also makes part of the coding simpler. I can easily patch by checking indexes when contracting (due to the priming).
I actually tried to have something a bit more generic than evaluating the order, but I was struggling a bit with reading the SmallStrings, and therefore postponed it.
I think it is important to well-differentiate between the nsites of the MPS (the period of the motif), the support of the Hamiltonian and the range.
Happy to take any convention you judge better.

src/vumps_localham.jl
```diff
@@ -120,7 +168,7 @@ function left_environment(hᴸ, 𝕙ᴸ, ψ; tol=1e-15)
   # Get the rest of the environments in the unit cell
   Hᴸ = InfiniteMPS(Vector{ITensor}(undef, N))
   Hᴸ[N] = Hᴸᴺ¹
-  Hᴸᴺ¹ = translatecell(Hᴸᴺ¹, -1)
+  #Hᴸᴺ¹ = translatecell(Hᴸᴺ¹, -1) #Found it was useless?
```
mtfishman, line 171: Good point, not sure why that was there. It doesn't look like `Hᴸᴺ¹` is used again after this, probably left over from an old version of the code.

src/vumps_localham.jl
```diff
@@ -19,35 +19,63 @@ struct Hᴬᶜ
...
-  δˡ⁻¹ = δ(l[n - 1], l′[n - 1])
   δʳ = δ(r[n], r′[n])
-  δʳ⁺¹ = δ(r[n + 1], r′[n + 1])
   Hᶜᴸv = v * Hᴸ[n] * dag(δʳ)
   Hᶜᴿv = v * δˡ * Hᴿ[n]
-  Hᶜʰv =
-    v * ψ.AL[n] * δˡ⁻¹ * ψ′.AL[n] * ∑h[(n, n + 1)] * ψ.AR[n + 1] * dag(δʳ⁺¹) * ψ′.AR[n + 1]
+  # TODO do we prefer a formalism like the commented one? if yes, need to define contraction of ITensor and MPS
+  #Hᶜʰv = v * ψ.AL[n] * δ(l[n - 1], l′[n - 1]) * ψ′.AL[n] * ∑h[(n, n + 1)] *
+  #  ψ.AR[n+1:n + range_∑h - 1]... * ψ′.AR[n+1:n + range_∑h - 1]... * dag(δ(r[n+range_∑h - 1], r′[n+range_∑h - 1]))
```
mtfishman, line 47: I guess that kind of notation could be nice, but we would have to be careful about proper order of operations for efficiency. I think the code below is reasonable, and we could consider more clever notation later.

src/vumps_localham.jl
```diff
@@ -19,35 +19,63 @@ struct Hᴬᶜ
...
+    temp_Hᶜʰv = ψ.AL[n-j] * δ(l[n - 1 - j], l′[n - 1 - j]) * ψ′.AL[n - j] * ∑h[(n - j, n + 1 - j)];
+    for k in 1:j
+        temp_Hᶜʰv  =  temp_Hᶜʰv  * ψ.AL[n - j + k] * ψ′.AL[n - j + k]
+    end
+    temp_Hᶜʰv = temp_Hᶜʰv  * v
+    for k in j+1:range_∑h-2
+        temp_Hᶜʰv =  temp_Hᶜʰv * ψ.AR[n - j + k] * ψ′.AR[n - j + k]
+    end
+    temp_Hᶜʰv =  temp_Hᶜʰv * ψ.AR[n - j + range_∑h - 1] * dag(δ(r[n - j +  range_∑h - 1], r′[n - j +  range_∑h - 1])) * ψ′.AR[n - j + range_∑h - 1]
+    Hᶜʰv = Hᶜʰv + temp_Hᶜʰv
```
mtfishman, line 63: I'll have to take your word that this is all correct (very easy to make off-by-one errors in code like this, as I'm sure you know), but it looks reasonable and doesn't seem overly complicated. Could you add a test of this functionality in some way? Even just a full VUMPS run with some longer range interactions where either you know the exact result or reproduce an energy that you think is correct.

mtfishman: Also obviously it is a good sign if the 2-local version of VUMPS is not affected by the addition of this code.

src/vumps_localham.jl
```diff
@@ -167,14 +215,13 @@ function (A::Aᴿ)(x)
   return xT - xR
 end
 
-function right_environment(hᴿ, ψ; tol=1e-15)
+# TODO Generate all environments, why? Only one is needed in the sequential version
```
mtfishman, line 218: It should be ok to only compute the environments that are needed (in practice there are a few needed, since you need the environments for `AC` and the bond matrices `C` on either side of `AC`). I think this was just leftover from copying code over from the parallel version.

However, in practice only one of the environments is computed using a linear solver (`linsolve`) and the rest are derived from that one through transfer matrices, so it should be relatively cheap to just compute all of them. We could revisit this if it actually leads to a noticeable performance problem.

src/vumps_localham.jl
```diff
@@ -297,15 +371,21 @@ function vumps_iteration_sequential(
...
-    Ãᴸ[n] = ortho_polar(Ãᶜ[n], C̃[n])
-    Ãᴿ[n] = ortho_polar(Ãᶜ[n], C̃[n - 1])
+    function ortho_polar_right(AC, C)
+      UAC, _ = polar(dag(AC), uniqueinds(AC, C))
+      UC, _ = polar(dag(C), commoninds(C, AC))
+      return noprime(dag(UAC)) * noprime(UC)
+    end
 
+    Ãᴸ[n] = ortho_polar(Ãᶜ[n], C̃[n])
+    Ãᴿ[n] = ortho_polar_right(Ãᶜ[n], C̃[n - 1])
```
mtfishman, line 381: Did the previous code cause a problem that you saw? I preferred trying to just have one function, but I don't think I had convinced myself that the previous code was properly equivalent to a right polar decomposition. 

The previous VUMPS code is working to very high accuracy in practice so I presume the previous version was correct.

mtfishman:
Thanks a lot for taking a look at this @LHerviou, this is some technical and tedious code to implement. This will be really useful for a wide variety of applications (FQHE, quantum chemistry, quasi-1d electron and spin models, etc.).

What is the status of this now? What types of interactions does this handle, and have you confirmed that it actually works on nontrivial long-range interacting Hamiltonians?

src/infinitemps.jl
mtfishman: Based on the proposal in #31 of using an `InfiniteMPOSum` representation, the MPOs could be constrained to be contiguous, in which case the range and support would be the same, but I think it is still useful to distinguish those two concepts.

For example, we should probably support allowing someone to input a long-range interacting Hamiltonian with terms that have support on two sites in terms of a 4-index tensor with non-contiguous sites (which we might then convert into an `InfiniteMPOSum` with contiguous sites equal to the range of the interaction).

Here is a code for finding the support of an operator with potentially non-contiguous sites:
```julia
using ITensors
using ITensorInfiniteMPS

ITensorInfiniteMPS.getcell(i::Index) = ITensorInfiniteMPS.getcell(tags(i))
function ITensors.findfirstsiteind(ψ::InfiniteMPS, i::Index)
  c = ITensorInfiniteMPS.getcell(i)
  i1 = translatecell(i, -(c - 1))
  n1 = findfirst(hascommoninds(i1), ψ[Cell(1)])
  return (c - 1) * nsites(ψ) + n1
end
function ITensors.findsites(ψ::InfiniteMPS, T::ITensor)
  s = filterinds(T; plev=0)
  return [ITensors.findfirstsiteind(ψ, i) for i in s]
end
ITensors.findsites(ψ::InfiniteCanonicalMPS, T::ITensor) = findsites(ψ.AL, T)

N = 2
heisenberg_space_shift(q̃nf, q̃sz) = [QN("Sz", 1 - q̃sz) => 1, QN("Sz", -1 - q̃sz) => 1]
heisenberg_space = fill(heisenberg_space_shift(1, 0), N)
s = infsiteinds("S=1/2", N; space=heisenberg_space)
initstate(n) = isodd(n) ? "↑" : "↓"
ψ = InfMPS(s, initstate)
model = Model("heisenberg")
H = InfiniteITensorSum(model, s)

@show findsites(ψ, H[(1, 2)])
@show findsites(ψ, H[(2, 3)])
@show findsites(ψ, H[(3, 4)])
@show findsites(ψ, H[(4, 5)])
```
This is following analogous functionality in `ITensors.jl` for finite MPS.

LHerviou: Even for an MPO representation of the local Hamiltonian, we do not need the MPO to have contiguous sites (putting nothing and waiting for contraction is equivalent to putting identity matrices in the MPO).

Thank you for the findsites template, I updated the PR, and added some matching functionalities for other situations.

src/vumps_localham.jl
LHerviou: I will check that tomorrow again, I think your code relied on the fact that you had two site per unit cell + some real terms.
I might be wrong though.

I agree that having one function would be best.

LHerviou:
Updated as discussed all functions related to the vumps_localham_sequential to work with multisites local Hamiltonians.
For optimization (in particular memory)

LHerviou:
Sorry for the misclick...

Currently the code is working perfectly fine for the next nearest neighbour Ising model I introduced.
This whether the unitcell of the MPS is taken to be 2, 3 or 4.
As mentioned somwhere else, I have some difficulties with the XX + YY +XZX +YZY, but it might be entanglement related (going from c=1 to c=2 is not cheap).

In the TODO list for very soon:
- improve the code for vumps_localham_sequential, taking into account the new findsites and nrange functions. This should make the code both simpler to read and more explicit. I will also add comments. I also want to have a quick look to the error measurements. It might need minor adjustments.
- think how to have site dependent local hamiltonian (the current constructor prevent it), such that one can study for example systems with alternative hoppings, without breaking the current generator.
- go towards the MPO representation of the local Hamiltonian. I am still not one hundred percent sure always decomposing is better than keeping local Hamiltonians.

src/infinitemps.jl
mtfishman: Right, the MPO does not need to be contiguous. Perhaps it makes it simpler to make it contiguous for coding (in terms of prime level logic) in which case you can simply write a function that puts identity tensors between non-contiguous sites of the MPO.

src/vumps_localham.jl
mtfishman: Ok. We have also seen some convergence issues with gapless models. At larger bond dimensions, we've seen the precision error fluctuating to order 1, which in practice can be improved by using a higher precision in the eigensolver. For example, you can try setting the new keyword argument `eigsolve_tol` introduced in #30 to `eigsolve_tol = (x -> x / 1000)` instead of the default `eigsolve_tol = (x -> x / 100)`. Not sure if that is related to the issue you are seeing though.

mtfishman:
Great to hear you already have that much functionality working! We would have immediate applications for the current functionality. For example @hershsingh is studying a Heisenberg model on a ladder which maps to a model with next nearest neighbor interactions, which he is currently handling by course graining to a doubled local Hilbert space with more complicated local interactions.

I'll leave it up to you, but if it is in a good place now, we could merge this once it is cleaned up a bit and leave things like interactions with non-contiguous sites and using an MPO representation of the local Hamiltonian terms for a future PR.

src/vumps_localham.jl
LHerviou: You were right, the previous code was working. Got confused by the indices.

LHerviou:
Updated with what we discussed. Tests should? be working now, including the formatting.
The previous regression on the Ising test came from the non-definition of OpSum in the corresponding model. I added it. Hopefully it does not break anything else used by other people (basically, the definition of a ITensor is not useful anymore in the models, as was implied/wished by the comments someone left).

For the rest, I included the possibility of non-contiguous terms and clarified the vumps_localham code. I include also some examples.
For my todo list:
- I want to investigate a bit the subspace expansion code. I am a bit surprised by some results, the code might need some additional modification. That is the only thing that might prevent the fusing as far as I am concerned. It is anyway unchanged for 2-local models.
- I will finally make a branch to work on the mpo representation of the state, as discussed.

mtfishman:
Sounds good. Could you please copy the extended Ising model example to the test suite to make a test out of it? You can just check the energy is correct to some precision. After the test is added, I'm good with merging this.

We can leave the MPO representation for a future PR. Probably this version is sufficient for most use cases anyway.

Interesting to hear that you suspect an issue in the subspace expansion code. What issues are you seeing?

LHerviou:
I am correcting exams, so the subspace expansion will be delayed. I included the test for the Ising Extended model.

Concerning the subspace expansion, the current formula simply does not correspond to a projection on the tangent space for Hamiltonians which are not two locals.

mtfishman:
Thanks for adding the test. I understand the problem, this expression:

https://github.com/ITensor/ITensorInfiniteMPS.jl/blob/e126828f990d1ffa518405d7a94738c2ab0a2201/src/subspace_expansion.jl#L34

would need to include all of the Hamiltonian terms that overlap with the sites adjacent to the bond being expanded.

I wonder if we can just reuse some of the code you already wrote in `Base.:*(H::Hᴬᶜ, v::ITensor)`? Essentially it would mostly be the same except extending to a two-site wavefunction and not applying the left and right environments, so most of the code could be shared and factored out into a separate function.

In fact, could we share more code across `Base.:*(H::Hᶜ, v::ITensor)` and `Base.:*(H::Hᴬᶜ, v::ITensor)` as well? Probably most of the functionality is the same. We could even just make a single core function for an n-site wavefunction, I imagine that is not too hard. If I remember correctly, Mike Zalatel once told me that it was important to use 3-site updates for iDMRG with FQHE because otherwise the optimization can get stuck, perhaps the same is true for VUMPS.

I'm ok leaving this issue (fixing the subspace expansion and reducing code duplication) for a future PR. I think it would be interesting to make this available, and presumably the subspace expansion being a bit off doesn't cause huge issues in general. But again, I'll leave that up to you.

LHerviou:
That should fix the error,I was trying to access functions which were not public.

The code can definitely be shared indeed. We can think about it at a later point.
I don't think that the subspace expansion being off is that important for now, but I willt try to fix it asap. At least for my peace of mind.

You are correct for the FQHE: as we have two U(1) symmetries,  (charge and momentum), the Hamiltonian is block diagonal over two sites. I am not yet sure how it would work in this case. Sadly, it might require more systematic use of subspace expansion to explore other blocks. I should be fixed soon.

test/test_vumps_extendedising.jl
```diff
@@ -62,15 +62,15 @@ using Random
   end
 
   function ITensors.expect(ψ::InfiniteCanonicalMPS, h::ITensor)
-    l = linkinds(only, ψ.AL)
-    r = linkinds(only, ψ.AR)
-    s = siteinds(only, ψ)
-    δˢ(n) = δ(dag(s[n]), prime(s[n]))
-    δˡ(n) = δ(l[n], prime(dag(l[n])))
-    δʳ(n) = δ(dag(r[n]), prime(r[n]))
+    l = linkinds(ITensorInfiniteMPS.only, ψ.AL)
```
mtfishman, line 65: This should be available in the namespace, it is exported by `Base`.

LHerviou:
It happens, that is the drawback of having such codes worked on by several people at the same time.
Let me know if I can help.

Btw, the upgrade to the MPO rep of the InfiniteTensorSum was mostly harmless and took me only two-three hours (at least for the core functions, and for the different tests to work).

LHerviou:
In particular, I can try to adapt the parallel algorithm so that it works accordingly. I did not do that yet, and that might be required for a peaceful integration

mtfishman:
Yes, we will need to figure out coordinating merging this with #37.

I think it is best to just go ahead and merge #37, and then we can fix any issues that might come up from that in this PR.

mtfishman:
@LHerviou I've pushed a few commits that fix the issue I introduced from merging the code from #38, and also merged the code from #37. Anything else that needs to be done for this PR?

mtfishman:
@LHerviou the parallel VUMPS code needed more adjustments for long-range interactions. I think I've fixed it. I've added a test for the parallel VUMPS code (though it just tests it for 2-local interactions).

LHerviou:
I added the parallel test for multisites. I read the code, everything seems good to me.
If it works as on my machine, I see no reason not to merge.

As soon as it is done, I will do the PR for replacing InfiniteTensorSum{ITensor} by a MPO

mtfishman:
Great, thanks again for such an involved and important contribution to the code! Very much appreciated, I think a lot of people will find this to be immediately useful.

I’ll merge once tests pass.

PR #44, opened 2022-02-03
Fixing the N=1 subspace expansion

LHerviou:
Here is the fix.

The code can probably be optimised, but it should be perfectly working now.
I needed it to benchmark vumps_mpo anyway.

Sorry about the random useless commits (the previous one). Apparently, the rebase was not done automatically.
I will try to do that before the next PR.


LHerviou:
Playing around with it, I think the error estimate might be off.

test/test_vumps.jl
```diff
@@ -38,15 +38,15 @@ using Random
...
-      # Flux density is inconsistent for odd unit cells
-      isodd(nsites) && continue
-    end
+    #if conserve_qns
+    # Flux density is inconsistent for odd unit cells
+    #  isodd(nsites) && continue
+    #end
 
-    space_ = fill(space_shifted(model, 0; conserve_qns=conserve_qns), nsites)
+    space_ = fill(space_shifted(model, 1; conserve_qns=conserve_qns), nsites)
```
mtfishman, line 49: Shouldn't it be `0` for even-site unit cells and `1` for odd-site unit cells?

mtfishman:
It seems like the code could be shared/compressed between the `N = 1` and `N > 1` cases, but I'm ok with dealing with that later if you don't have time to look into that right now. I think it is more important for this to be working so we can test it out.

test/test_vumps.jl
LHerviou: Due to the 2-modularity, for the chosen initial state, anything goes when the number of sites is even..

LHerviou:
I agree, there is definitely some overlap. I just wanted a quick and dirty fix to be able to work on the vumps_mpo.

Building on what you did in vumps_localham, I have something that appears to be working for nsites(\psi ) = 1

test/test_vumps.jl
mtfishman: Ah I see, thanks for the explanation.

mtfishman:
So should we go ahead and merge this?

LHerviou:
I think so.
Please let me know if any other problems appear.

Once it is done and I manage to clean the history, I will do a PR for the MPO ref of the InfiniteTensorSum, and then hopefully of the MatrixMPO representation.

mtfishman:
Great, thanks!

PR #45, opened 2022-02-08
Replacing the local ITensors by MPOs in InfiniteITensorSum

LHerviou:
Replace the ITensor in InfiniteITensorSum by a MPO.
The change is completely transparent from the user point of view.

Improved the tests on the extended Ising (slows down the testing quite significantly).

Improve the shared code in vumps_localham.jl

I might be interesting to check there are no significant performance regression. It did not seem to be the case in my testing.

### DIFF
--- a/src/ITensors.jl
+++ b/src/ITensors.jl
@@ -72,6 +72,10 @@ function Base.startswith(tag::Tag, subtag::Tag)
 end
 
 function tag_starting_with(ts::TagSet, prefix)
+  x = findfirst(t -> startswith(t, Tag(prefix)), ts)
+  if isnothing(x) #in case the function is called one a link leg. Can be probably improved
+    return x
+  end
   return ts[findfirst(t -> startswith(t, Tag(prefix)), ts)]
 end
 
--- a/src/infinitemps.jl
+++ b/src/infinitemps.jl
@@ -57,14 +57,24 @@ end
 
 ITensors.siteinds(f::typeof(only), ψ::InfiniteCanonicalMPS) = siteinds(f, ψ.AL)
 
-ITensorInfiniteMPS.getcell(i::Index) = ITensorInfiniteMPS.getcell(tags(i))
+getcell(i::Index) = ITensorInfiniteMPS.getcell(tags(i))
 getsite(i::Index) = getsite(tags(i))
+
+#Before, for a two site Hamiltonian, findsites(ψ, H[3]) would return [1, 2]
+#Appeared unsafe for index purpose
 function ITensors.findfirstsiteind(ψ::InfiniteMPS, i::Index)
   c = ITensorInfiniteMPS.getcell(i)
-  i1 = translatecell(i, -(c - 1))
-  n1 = findfirst(hascommoninds(i1), ψ[Cell(1)])
+  n1 = getsite(i)
   return (c - 1) * nsites(ψ) + n1
 end
+
+function ITensors.findsites(ψ::InfiniteMPS, T::MPO)
+  s = [noprime(filterinds(T[x]; plev=1)[1]) for x in 1:length(T)]
+  return sort([ITensors.findfirstsiteind(ψ, i) for i in s])
+end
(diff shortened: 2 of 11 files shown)


src/infinitemps.jl
```diff
@@ -74,38 +84,57 @@ ITensors.findsites(ψ::InfiniteCanonicalMPS, T::ITensor) = findsites(ψ.AL, T)
 # For now, only represents nearest neighbor interactions
 # on a linear chain
 struct InfiniteITensorSum
```
mtfishman, line 86: Maybe we should rename this to `InfiniteSum{T}`, then define `InfiniteITensorSum = InfiniteSum{ITensor}` and `InfiniteMPOSum = InfiniteSum{MPO}`?

mtfishman: Then most code could be written in terms of `InfiniteSum`, for example, and certain methods could be specialized to the ITensor or MPO version.

mtfishman:
Looks really great over all @LHerviou! Looks like changing to the MPO representation didn't change much of the environment calculation code itself, which is nice. Also great to see the code refactoring to reduce code duplication across the serial and parallel versions, the code is much more readable and will be a lot easier to maintain, and easier to extend to other Hamiltonian representations like MPOs.

Is the current plan for fermions that the MPO terms will just need Jordan-Wigner strings in them?

src/ITensors.jl
```diff
@@ -72,6 +72,10 @@ function Base.startswith(tag::Tag, subtag::Tag)
 end
 
 function tag_starting_with(ts::TagSet, prefix)
+  x = findfirst(t -> startswith(t, Tag(prefix)), ts)
+  if isnothing(x) #in case the function is called one a link leg. Can be probably improved
```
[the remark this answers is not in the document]
LHerviou: Well, when I am using the MPO representation of the Hamiltonians, the inner links are not labelled with $c=...$, which lead to some problems.

This was an easy fix.

src/infinitemps.jl
LHerviou: I think it is indeed the best way to do it. InfiniteSum{ } can be done easily and the only changes will be in contractions.

LHerviou:
Indeed, the representation in MPO simplifies  some considerations, but the computation itself is mostly left unchanged (just some additional tensors present sometimes).

I must say I did not think too much about fermions for now. My initial idea was indeed to have explicit Jordan-Wigner strings in them. How does it work exactly by default in ITensors? When I build a MPO on a fermion lattice, does it store explicitely the string in the MPO rep or does the magic spontaneously happens when contracting?

src/infinitemps.jl
LHerviou: Do you want me to have an attempt at it?

mtfishman: Yeah, I could do that.

mtfishman:
> I must say I did not think too much about fermions for now. My initial idea was indeed to have explicit Jordan-Wigner strings in them. How does it work exactly by default in ITensors? When I build a MPO on a fermion lattice, does it store explicitely the string in the MPO rep or does the magic spontaneously happens when contracting?

The situation right now is a bit complicated. Basically, there are three different ways fermions are handled right now:

1. Jordan-Wigner string: This gets inserted automatically in the OpSum to MPO constructor in ITensors.jl, as well as in functions like `correlation_matrix`.
2. Fermionic swap gates: This is how fermions are handled in the `apply` function for doing gate evolution of an MPS.
3. Automated fermion sign: This is an experimental feature enabled with `ITensors.enable_auto_fermion()`, with the goal that it automatically handles the fermion sign in every basic operation in ITensor (contraction, permutation, addition, SVD, etc.). The plan is for that to supersede 1. and 2., but there are some bugs to work out in trickier operations like SVD so it is only experimental for now.

For this case, we can just handle the fermion sign in an analogous way to 1. and encode it with a Jordan-Wigner string in the MPO.

PR #50, opened 2022-02-09
Vumps with Infinite MPO

LHerviou:
The big update with Vumps working for an Hamiltonian represented as an infinite MPO.

As discussed, for now, we represent the infinite MPO as a matrix of ITensor automatically built from the MPO representation of the local Hamiltonian of the models. It assumes the MPO is triangular inferior.
This has the advantage of removing some edge cases in the computation of the environments.
I include also a test file.

Not that as of now, it does not support infinite range Hamiltonians (with exponentially decaying terms). I might do a PR another day to take this into account, but it is a few lines of code to fix it.

Some additional asserts might be required to make it fool/future proof (I do not properly check that the diagonal terms of the MPO are identity matrices exactly, and that the Matrix is triangular inferior). I will see if I can find the time to do that today or tomorrow.

The generic cell translation will wait for another PR, but is on the way. It required some small rewrites pretty much everywhere so I prefer to keep it separated.


src/vumps_mpo.jl
```diff
@@ -0,0 +1,540 @@
...
+# `src/vumps_localham.jl`. Help getting this working would be highly
+# appreciated so we can support more generic Hamiltonians.
+# Ideally, the implementation would only overload the minimal amount
+# of code needed to be modified (like computing quasi-left and
+# right environments and updates to `AC` and `C`), and share the high
+# level code with the existing VUMPS code in `src/vumps_localham.jl`.
+#
+# Please reach out on this issue: https://github.com/ITensor/ITensorInfiniteMPS.jl/issues/39
+# if you are interested in implementing this feature.
+#################################################################
```
mtfishman, line 15: Happy to see this comment go!

mtfishman:
This is great, thanks a lot for getting started on this. Over all the code looks well written.

I'm surprised at how simple the `InfiniteMPOMatrix` turned out to be! I didn't realize that it would be so simple if you start from the `InfiniteSum{MPO}` representation, I guess I never thought hard about how that type of conversion would work but it seems like you basically just map the sites that the terms of the `InfiniteSum{MPO}` act on to matrix elements of the `InfiniteMPOMatrix` representation?

In principle I was hoping that there would be a single generic version of both `tdvp_iteration_sequential` and `tdvp_iteration_parallel` which accepted different forms of the Hamiltonian (so for example `InfiniteSum{MPO}`, `InfiniteMPOMatrix`, and eventually `InfiniteMPO`), and that the only parts that need to be customized are functions like `left_environment`, `right_environment`, and the definitions of the 0-site and 1-site effective Hamiltonians. Is that the design you ultimately had in mind?

LHerviou:
Thanks a lot, but having the CellVectors infrastructure and a good base for the code was extremely helpful.

About the `InfiniteMPOMatrix`: yes, it is basically just that. I figured it out a while ago while playing with/explaining to one student the graph construction of MPOs. Instead of fully separating the `InfiniteTensorSum` into simple tensor products, we can condense together all steps going from j to j+1 using a tensor network representation.
It is of course non-optimal at all, but it is very very simple. And it can be obviously generalized to long ranges if need be by threading identity tensors.

It is definitely possible to do that now, I think. The codes have a very similar structure. I do not know how much gain/simplicity it brings though, given that there are some subtleties in the definition of the environments.

mtfishman:
> About the `InfiniteMPOMatrix`: yes, it is basically just that. I figured it out a while ago while playing with/explaining to one student the graph construction of MPOs. Instead of fully separating the `InfiniteTensorSum` into simple tensor products, we can condense together all steps going from j to j+1 using a tensor network representation. It is of course non-optimal at all, but it is very very simple. And it can be obviously generalized to long ranges if need be by threading identity tensors.

Very cool, I'll have to think about that a bit. Nice to have a simple construction as a reference. Also we could use https://arxiv.org/abs/1909.06341 as a post-processing compression.
 
> It is definitely possible to do that now, I think. The codes have a very similar structure. I do not know how much gain/simplicity it brings though, given that there are some subtleties in the definition of the environments.

I think the advantage is in terms of code maintenance and extensibility. Two situations that might come up are adding some new logic to the TDVP/VUMPS iteration code, such as adding a new keyword argument that doesn't affect the deeper logic of the `left_environment`/`right_environment` code directly. If the code is merged into a single generic implementation, you don't have to keep the multiple versions up to date. Additionally, it makes it easier to implement a new Hamiltonian representation (for example, I would want to implement code for VUMPS with sums of infinite MPOs which can lead to simple parallelization, which would share most of the outer logic of the TDVP update).

LHerviou:
Need a delay before integrating. I found a bug this morning. Will let you know.

LHerviou:
Fixed a bug that occurred for larger Matrix{MPO}. 

Added the relevant tests.

LHerviou:
Sorry for the long delay, holidays followed by covid.
Fixed the integration of the new InfiniteSum, including two small conversions between InfiniteSum types.

mtfishman:
No worries, sorry to hear about getting COVID and hope all is well. Sorry for the delay on my end, I was traveling and am now catching up on reviewing PRs.

Over all this looks good to me. I haven't looked through all of the code in detail, but the style looks reasonable and I presume it is correct from the tests you've added. I would still like to make generic versions of `tdvp_iteration_sequential` and `tdvp_iteration_parallel` that are shared by the different MPO representations we now have, but I can add that in a future PR.

Anything else you would like to add, or should I go ahead and merge?

LHerviou:
No worries for the delay, I keep working on my code anyway.

I think you can merge. I will be busy this week, but the next I will try to make a PR with the generalized translation.

PR #53, opened 2022-03-22
Translation added

LHerviou:
Hi,

Here are the modifications that allow to have generalized translation.
I tested that everything seem to work (including on fqhe type of translations, examples not included here). 
It is nonetheless not impossible that I missed something. There are a lot of small functions that assume some basic properties on the translation operator.

I am available for discussions, and of course all names can be changed.


src/infinitemps.jl
```diff
@@ -65,9 +67,44 @@ getsite(i::Index) = getsite(tags(i))
 function ITensors.findfirstsiteind(ψ::InfiniteMPS, i::Index)
   c = ITensorInfiniteMPS.getcell(i)
   n1 = getsite(i)
-  return (c - 1) * nsites(ψ) + n1
+  if translater(ψ) == translatecell
+    return (c - 1) * nsites(ψ) + n1
+  else
```
mtfishman, line 72: Could you explain this section of code?

I guess you are looping through the indices and seeing which site matches, up to some limit of 1000 sites? I remember thinking I would have to design it that way until I thought of just using the tag but I could see that is no longer possible with a general translation function.

Is there some assumption about the relationship between the tags and the `translater` function? Since it still looks like you are making use of `getcell` and `getsite` in the general case.

mtfishman:
Thanks for this! Definitely a pretty involved change to make, I guess it was quite ingrained in the code.

Glad to hear it works for FQHE-type calculations. We are interested in doing DMRG and VUMPS calculations on systems like that so I would definitely be interested in hearing about your experience and it's very helpful that you are pushing the code in that direction!

I have two comments about the notation:

1. What about `translator` instead of `translater`, since `translater` isn't a real word? I was considering something like `translation_function` but `translator` is nice and succinct.
2. I noticed the default function is still called `translatecell`. Maybe to distinguish it more from the general case we could call it `translatecelltags`? That emphasizes that it only translates the tags, instead of something more general. I'm talking about functions like:
```julia
function translatecell(ts::TagSet, n::Integer)
  ncell = getcell(ts)
  if isnothing(ncell)
    return ts
  end
  return replacetags(ts, celltags(ncell) => celltags(ncell + n))
end

function translatecell(i::Index, n::Integer)
  ts = tags(i)
  translated_ts = translatecell(ts, n)
  return replacetags(i, ts => translated_ts)
end
```

src/infinitemps.jl
LHerviou: Yes, the idea is that this way one can deal with more general tags and it allows for an easy benchmarking: namely replacing translation c->c+1 by c->c+2.

The original version assumed that for the tags, translation by one unit cell is c->c+1
Here we just assume that c is an increasing function.

I am not sure it is actually a useful feature to be honest. It might be worth putting the extension in a test, and just specifying that the tag pattern for translation must remain the same. (In fact thinking about it, it is probably the best solution, I will fix it like that if you agree).

I updated the getsite function (or at least the getsite) to be a bit more generic in order to treat this kind of edge states. It can be rolled back easily.

LHerviou:
There was a lot of little changes to make. Nothing too bad, but it is easy to miss things.

1 -  Concerning the name: seems fine by me. Translater is a bastardized frenchified version, as translator translates in a different way. I will update that.
2 - Probably indeed better to avoid confusion on some pieces of the code.


So far, due to the symmetries (two site updates and expansion are not enough to be "ergodic"), I have some difficulties making VUMPS work on FQHE states. More precisely, there are still some issues with the subspace expansion steps.
I implemented a working adaptable iDMRG that is able to treat the FQHE states, and I will go back to VUMPS in a few days using this code as a basis to deal with dangling legs. In particular the unitarity constraint is harder to satisfy (one wants to add too many legs in order to guarantee a successful expansion), so having a few steps of n-site iDMRG at the beginning just to leave the product-state "manifold" might be required.

Except from that, the code work with the symmetries of the FQHE, so a temporary solution would be alternating iDMRG steps to generate the correct MPS links, and VUMPS steps for convergence.

Did you have any plan to include iDMRG codes in the package?

src/infinitemps.jl
mtfishman: Yeah, it sounds reasonable to assume the `translator` functions shifts the cell tags by 1 for now, and if we come across a more general case than we can generalize it later.

mtfishman:
Interesting, yes I was wondering if the current subspace expansion would work with FQHE Hamiltonians.

I remember hearing that you need to use a 3-site update in DMRG/iDMRG for FQHE Hamiltonians or else it gets stuck ("non-ergodic" as you put it). Could it just be a matter of generalizing the subspace expansion to 3-sites? If so, that would be a relatively easy fix.

I would consider adding iDMRG, but I was hoping that VUMPS would be sufficient for most, if not all, use cases (with the advantage that it should be faster in general). In general it would be nicer to just maintain one algorithm/code. Of course, iDMRG is better tested and more widely used in general, with a long history of improvements to the algorithm, so VUMPS is less "battle tested" and may have corner cases like the FQHE that need to be worked out. However, if you already have a working version of iDMRG, it would be great to have, at the very least as a reference to check the performance and accuracy of VUMPS.

src/infinitemps.jl
LHerviou: Sure. I will modify the code.

LHerviou:
It does not work, and it is pretty simple to see: if you start from a product state, the non-physical legs of the two-site Hamiltonian entirely fix all quantum number within the two sites, such that no new subspace is generated/the local Hamiltonian is block-diagonal.

From what I have seen, even three sites can get stuck easily. I do not have any proof of that, but as an example, I have worked on a minimal model in a completely different context of the form c^d_1 c_2 c_3 c^d_4 (the first hopping which respect the FQHE symmetry). This model undergoes Hilbert space fragmentation.
In the 2012 paper of Mike Zaletel and Frank Pollmann, they had to go to 6 to 8-site DMRG for complex states (Pfaffian, 1/5, 2/5 etc...).

On paper it should not be too hard, but it seems to have some instabilities that I have not be fully able to solve. I need to go back to it. 
In particular, left/right unitarity is extremely easy to break, especially if one goes to larger cylinders/bigger spheres. If it is broken, then VUMPS does not converge well.

My iDMRG code is pretty non-efficient (it has some "useless" features for my use case/that I did not bother improving) , but I can add it at some point for debugging purposes. The FQHE is also a good test bed for mixers/noise generators.

mtfishman:
Yes, I think there are some interesting questions here, and it is an open research problem to come up with one good, efficient subspace expansion technique that can cover many cases, especially these tricky ones like FQHE. It may end up that there are tradeoffs in terms of performance and reliability and we will need to provide a variety of options which users will need to try in challenging cases. Having iTEBD and iDMRG implementations could fall into that category, for example I've heard a good way to run TDVP is by first running TEBD for some time to expand the bond dimension and then switch to TDVP (for example that is what is used here: https://arxiv.org/pdf/2009.10736.pdf). There are also global Krylov methods people have tried (https://arxiv.org/pdf/2005.06104.pdf) which could be adapted to the infinite case.

Interesting to hear about the polar decomposition issue. We can also use a positive QR decomposition (a QR decomposition with the diagonal of R fixed to positive) as an alternative if the polar fails.

By "mixer" do you mean the subspace expansion technique introduced in: https://arxiv.org/abs/1501.05504? I was wondering if that technique could be adapted to the translational invariant case but I didn't figure out how to do it.

It seems like the easiest thing to try for now is generalizing the current subspace expansion method in the VUMPS code to n-sites and see if that fixes most of the issues, and then investigate other options if not.

mtfishman:
I guess the major change were #55 and #56.

There are now 3 versions of VUMPS:

1. `InfiniteSum{ITensor}`: The old version of VUMPS in terms of an infinite sum of ITensors, only works with 2-local Hamiltonians. Brought back in #55 mostly to try to debug the performance issues of `InfiniteSum{MPO}` but we probably should get rid of it again once we are sure `InfiniteSum{MPO}` works well.
2. `InfiniteSum{MPO}`: Your general version of VUMPS in terms of an infinite sum of MPOs. We saw performance issues with this version, but they were fixed in #56.
3.`InfiniteMPOMatrix`: Your version of VUMPS written in terms of a single infinite MPO.

Which version are you seeing issues with? We have someone who I believe is using versions 1. and 2. and hasn't reported any issues, but he is just using a 2-site unit cell with a 2-local Hamiltonian.

LHerviou:
I was indeed talking about the third version.
Maybe a change in the dependencies then.

The test in the norm AL AL...C - C AR AR... is no longer passing.

LHerviou:
By the way, could you specify the exact test you did to check for the regression ?
I guess it is part of the file examples/vumps/vumps_hubbard_extended.jl ? Or the full file?

mtfishman:
> EDIT: I fixed the problem, it was my fault mainly.
> 
> Do you want me to reinstate the multisite version of VUMPS using local Hamiltonians accessible [here](https://github.com/LHerviou/ITensorInfiniteMPS.jl/blob/76c32a236152960fd2f4fd9455cfa5de746a7661/src/vumps_localham.jl)
> 
> We can put an if close such that it uses the current version when the Hamiltonian is two-local

Glad you figured out the problem.

I think I would rather just drop version 1. (`InfiniteSum{ITensor}`) entirely if possible in favor of 2. (`InfiniteSum{MPO}`) and 3. (`InfiniteMPOMatrix`), at least eventually. Is there any advantage of the version you link to above over the current `InfiniteSum{MPO}` we have already? If not I would say no, if anything we should be reducing the number of version of VUMPS that we have.

In fact I would hope that eventually `InfiniteMPO`/`InfiniteMPOMatrix` could supersede `InfiniteSum{MPO}` once we have a good AutoMPO-like system for infinite MPOs, like how for DMRG we only have an MPO version (though it is written generically so that in principle you can input any Hamiltonian representation that you want to).

mtfishman:
> By the way, could you specify the exact test you did to check for the regression ? I guess it is part of the file examples/vumps/vumps_hubbard_extended.jl ? Or the full file?

You mean the performance regression fixed in #56? That I was testing by running https://github.com/ITensor/ITensorInfiniteMPS.jl/blob/main/examples/vumps/vumps_hubbard_extended.jl.

LHerviou:
> Glad you figured out the problem.
> 
> I think I would rather just drop version 1. (`InfiniteSum{ITensor}`) entirely if possible in favor of 2. (`InfiniteSum{MPO}`) and 3. (`InfiniteMPOMatrix`), at least eventually. Is there any advantage of the version you link to above over the current `InfiniteSum{MPO}` we have already? If not I would say no, if anything we should be reducing the number of version of VUMPS that we have.
> 
> In fact I would hope that eventually `InfiniteMPO`/`InfiniteMPOMatrix` could supersede `InfiniteSum{MPO}` once we have a good AutoMPO-like system for infinite MPOs, like how for DMRG we only have an MPO version (though it is written generically so that in principle you can input any Hamiltonian representation that you want to).

Depending on the bond dimension of the MPO, I suspect the `InfiniteSum{ITensor}` is better than the `InfiniteSum{MPO}`. Not necessarily by much though.

mtfishman:
Yeah, I imagine the difference won't be too much. In #56 for 2-local interactions I saw the time was nearly the same. I imagine the MPO version should be better for non-local interactions, which as you know was the point of implementing it in the first place. I'm not sure if it is worth keeping around both implementations since as we see already it is a challenge supporting many different versions (each can have their own bugs, performance optimizations, etc.).

mtfishman:
@LHerviou is this ready to merge?

mtfishman:
Ok, sounds good! Is the main reason it would not be fully functional because of the subspace expansion?

LHerviou:
Exactly.

I had a prototype for the subspace expansion, but it was still failing. Following some playing with iDMRG (the subspace expansion is very close to doing one step of iDMRG after all), I might have an idea to help with that.

From memory, one of the key problems was that after expansion, AL... C - C AR... start to be very far from 0.
Another limitation was that it was hard (i.e. required really large bond dimensions and steps) to get a non-trivial, coherent expansion.

mtfishman:
Interesting, thanks for the update. There are definitely some open questions remaining about how to best do subspace expansion for a variety of problems, and FQHE sounds particularly challenging. We will be adding some subspace expansion techniques to our finite TDVP code (https://github.com/ITensor/ITensorTDVP.jl). One of them will be essentially the same as the one from this package (with the goal of sharing the implementation as much as possible), and we will also add an implementation of the technique from [this paper](https://arxiv.org/abs/2005.06104), however we will be exploring some new ones which would hopefully help with FQHE VUMPS as well.

Let me know how your experiments go, using iDMRG for subspace expansion sounds hopeful. Perhaps once your and our experiments get a bit farther along we can meet to discuss the different subspace expansion options. I'm hoping that our VUMPS/TDVP packages and the wider community can converge on one good subspace expansion technique that could work for a variety of techniques and problems (VUMPS, single-site DMRG, TDVP, FQHE, ladder systems, etc.) but that may be challenging since there will likely be tradeoffs between how robust the expansion is and how expensive it is.

LHerviou:
If the tests are passing as on my computer, this is ready to ship.

test/test_vumpsmpo_fqhe.jl
```diff
@@ -0,0 +1,98 @@
...
+      fermions_space_shift(
+        x; p=p, q=q, conserve_momentum=conserve_momentum, momentum_shift=momentum_shift
+      ) for x in 1:nsites
+    ]
+    s = infsiteinds(
+      "Fermion", nsites; space=fermionic_space, translator=fermion_momentum_translator
+    )
+    ψ = InfMPS(s, initstate)
+
+    Hmpo = InfiniteMPOMatrix(model, s, fermion_momentum_translator; model_kwargs...)
```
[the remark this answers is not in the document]
LHerviou: Oh, of course... I do it for the rest.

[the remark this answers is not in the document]
LHerviou: It is actually already implemented. I had just forgotten about it when quickly writing the additional test.

mtfishman:
Looks like the last test is taking a long time to run for some reason.

I'll go ahead and merge this. Thanks @LHerviou!

PR #55, opened 2022-04-05
Bring back `InfiniteSum{ITensor}` for 2-local Hamiltonians

mtfishman:
I've gotten reports from Bo Xiao of CCQ that there appears to be a performance/memory regression in the VUMPS/TDVP code, likely caused by the rewrite of the code for non-local interactions.

To help debug the performance issue, this PR brings back the simpler version of VUMPS/TDVP that was written for 2-local Hamiltonians in terms of sums of ITensors.

The interface is that if you input a `InfiniteSum{MPO}` it calls the newer version for non-local Hamiltonians (the version that appears to have a performance/memory regression) while inputting an `InfiniteSum{ITensor}` calls the old version that is limited to 2-local Hamiltonians.

@LHerviou have you noticed any performance issues since your rewrite of the VUMPS/TDVP code that extended it to non-local Hamiltonians?

### DIFF
--- a/examples/vumps/vumps_hubbard_extended.jl
+++ b/examples/vumps/vumps_hubbard_extended.jl
@@ -9,6 +9,7 @@ maxdim = 30 # Maximum bond dimension
 cutoff = 1e-6 # Singular value cutoff when increasing the bond dimension
 max_vumps_iters = 200 # Maximum number of iterations of the VUMPS algorithm at each bond dimension
 outer_iters = 5 # Number of times to increase the bond dimension
+localham_type = ITensor
 
 model_params = (t=1.0, U=10.0, V=0.0)
 
@@ -36,7 +37,7 @@ model = Model"hubbard"()
 @show model, model_params
 
 # Form the Hamiltonian
-H = InfiniteSum{MPO}(model, s; model_params...)
+H = InfiniteSum{localham_type}(model, s; model_params...)
 
 # Check translational invariance
 println("\nCheck translational invariance of initial infinite MPS")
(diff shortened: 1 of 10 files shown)


mtfishman:
I'm merging so we can test it out.

mtfishman:
I just did timings of the Hubbard model example, with these parameters:
```julia
maxdim = 30 # Maximum bond dimension
cutoff = 1e-6 # Singular value cutoff when increasing the bond dimension
max_vumps_iters = 200 # Maximum number of iterations of the VUMPS algorithm at each bond dimension
outer_iters = 5 # Number of times to increase the bond dimension
localham_type = MPO # or ITensor

model_params = (t=1.0, U=10.0, V=0.0)
```
The timing results I get for the old VUMPS implementation (`InfiniteSum{ITensor}`) are:
```julia
 21.846960 seconds (269.77 M allocations: 25.508 GiB, 18.09% gc time, 0.02% compilation time)
```
while for the new VUMPS implementation (`InfiniteSum{MPO}`) it gives:
```julia
 36.917972 seconds (473.90 M allocations: 41.024 GiB, 16.59% gc time, 0.01% compilation time)
```
The runs are essentially exactly the same, with the final energies equal to each other up to 15 digits.

I bet it is just a matter of a contraction being done in a non-optimal way somewhere (i.e. a non-optimal contraction sequence). I will try to investigate if I have some time, but at least now the the old implementation is available so we can use that in the meantime while we track down where the performance bug is.

LHerviou:
Just a simple question then:
if I write the code A = B * C * D ..., or A *= B * C * D, in which order is the contraction performed by ITensor?

mtfishman:
You can enable automatic optimization with `ITensors.enable_contraction_sequence_optimization()`, but we don't have it on by default since it has some overhead (and in general scales exponentially with the number of tensors in the contraction).

Glancing through the code, I do see some parts with non-optimal sequences so hopefully it is a simple fix of reordering some contractions.

mtfishman:
For example lines like this:

https://github.com/ITensor/ITensorInfiniteMPS.jl/blob/195bbf9880e8c78e0d43264bb28689c1243494af/src/vumps_nonlocalham.jl#L51

I assume should be:
```julia
        temp_Hᶜʰv = temp_Hᶜʰv * ψ.AL[n - j + k] * ∑h[n - j][idx] * ψ′.AL[n - j + k]
```

LHerviou:
That is precisely what I had in mind, yes.

PR #56, opened 2022-04-05
Optimize nonlocal Hamiltonian version of VUMPS

mtfishman:
Start to address the performance issue from #55.

Timings for the Hubbard model example, with these parameters:
```julia
maxdim = 30 # Maximum bond dimension
cutoff = 1e-6 # Singular value cutoff when increasing the bond dimension
max_vumps_iters = 200 # Maximum number of iterations of the VUMPS algorithm at each bond dimension
outer_iters = 5 # Number of times to increase the bond dimension
localham_type = MPO # or ITensor

model_params = (t=1.0, U=10.0, V=0.0)
```
The timing results I get for the old VUMPS implementation (`InfiniteSum{ITensor}`) are:
```julia
 21.846960 seconds (269.77 M allocations: 25.508 GiB, 18.09% gc time, 0.02% compilation time)
```
With this PR, the timing of the new VUMPS implementation (`InfiniteSum{MPO}`) gives:
```julia
 21.031708 seconds (296.70 M allocations: 28.086 GiB, 17.55% gc time, 0.11% compilation time)
```

Before this PR, the timing for the new VUMPS implementation (`InfiniteSum{MPO}`) was:
```julia
 36.917972 seconds (473.90 M allocations: 41.024 GiB, 16.59% gc time, 0.01% compilation time)
```

@LHerviou

### DIFF
--- a/src/vumps_nonlocalham.jl
+++ b/src/vumps_nonlocalham.jl
@@ -24,51 +24,51 @@ function Base.:*(H::Hᶜ{MPO}, v::ITensor)
   Hᶜᴿv = v * δˡ(n) * Hᴿ[n]
   #We now start building terms where C overlap with the local Hamiltonian
   # We start with the tensor AL[n] - v - AR[n+1] ... AR[n + range_∑h - 1]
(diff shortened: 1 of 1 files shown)


mtfishman:
@LHerviou there may be some other optimizations to make but the timing is very close to the old local Hamiltonian version now.

Issue #68, opened 2023-01-11
`infsiteinds` with QNs and arbitrary initial state


nbaldelli:
If I try to initialize an infiniteMPS with an arbitrary initial state, for example 

```
N = 2 # Unit cell size
initstate(n) = isodd(n) ? "↑" : "0" #quarter filling
s = infsiteinds("Electron", N; initstate, conserve_qns = true)
ψ = InfMPS(s, initstate)
```

I get this error

```
ERROR: Dividing ITensors.QNVal(Nf, 1, -1) by 2, the resulting QN value is not an integer
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] /(qnval::ITensors.QNVal, n::Int64)
    @ ITensorInfiniteMPS ~/.julia/packages/ITensorInfiniteMPS/Ogcvb/src/infinitecanonicalmps.jl:10
  [3] #103
    @ ~/.julia/packages/ITensorInfiniteMPS/Ogcvb/src/infinitecanonicalmps.jl:17 [inlined]
  [4] macro expansion
    @ ~/.julia/packages/StaticArrays/PUoe1/src/mapreduce.jl:75 [inlined]
  [5] _map
    @ ~/.julia/packages/StaticArrays/PUoe1/src/mapreduce.jl:40 [inlined]
  [6] map
    @ ~/.julia/packages/StaticArrays/PUoe1/src/mapreduce.jl:31 [inlined]
  [7] /
    @ ~/.julia/packages/ITensorInfiniteMPS/Ogcvb/src/infinitecanonicalmps.jl:17 [inlined]
  [8] shift_flux_to_zero(s::Vector{Index{Vector{Pair{QN, Int64}}}}, flux::QN)
    @ ITensorInfiniteMPS ~/.julia/packages/ITensorInfiniteMPS/Ogcvb/src/infinitecanonicalmps.jl:56
  [9] shift_flux_to_zero(s::Vector{Index{Vector{Pair{QN, Int64}}}}, initstate::Function)
    @ ITensorInfiniteMPS ~/.julia/packages/ITensorInfiniteMPS/Ogcvb/src/infinitecanonicalmps.jl:38
 [10] infsiteinds(site_tag::String, n::Int64; translator::Function, initstate::Function, kwargs::Base.Pairs{Symbol, Bool, Tuple{Symbol}, NamedTuple{(:conserve_qns,), Tuple{Bool}}})
[...]

PR #69, opened 2023-02-20
Generalize `infsiteinds` to fractional QN densities

LHerviou:
Hej Mat,

A minor update to fix the issue #68 
I am using it as an excuse to ask how to proceed with the following.

On my own fork, I have finished the implementations of a few useful features (paper coming out soon using it). This includes, from trivial to less trivial:
- SU(3) and SU(4) irrep using color conservation (nothing non-Abelian)
- some small fixes in general
- implementation of a few convenient functions for data analysis (QN-resolved analysis of ent spectrum, transfer matrix spectrum + some correlation functions)
- a working iDMRG for InfiniteMPOMatrix
- a working InfiniteMPO constructed from an InfiniteMPOMatrix
- a working iDMRG for InfiniteMPO
- compression of iMPOs using InfiniteMPOMatrix and the approach by Parker, Cao et Zaletel in [Phys.Rev.B 102](https://doi.org/10.1103/PhysRevB.102.035147) . It is not yet fully benchmarked so that should clearly be _experimental_ , especially as it is the simplest, non-fail proof version.
- I also plan to do this week some modifications on the subspace expansion as I cannot afford the default version in my new project (the one that required the compression of iMPOs)

It might be a lot for a single merge request. So how would you like to proceed? It will take me one or two weeks anyway to make things a bit cleaner.


LHerviou:
The failure does not seem to come from my code but from the new @require.

src/infinitecanonicalmps.jl
```diff
@@ -48,13 +60,28 @@ function shift_flux(i::Index, flux_density::QN)
...
 function shift_flux_to_zero(s::Vector{<:Index}, flux::QN)
   if iszero(flux)
     return s
   end
   n = length(s)
-  flux_density = flux / n
-  return map(sₙ -> shift_flux(sₙ, flux_density), s)
+  try
+    flux_density = flux / n
+    return map(sₙ -> shift_flux(sₙ, flux_density), s)
```
mtfishman, line 80: Could we explicitly check for what might go wrong to cause the error, instead of using `try ... catch ...`? The only reason I use `try ... catch ...` is to catch errors in external libraries that I have no control over, so I can fail gracefully or output a nicer error message (i.e. catching an error from LAPACK). Beyond that I think it is better to use `if` statements.

mtfishman:
Thanks for the fix @LHerviou! @nbaldelli, does this address #68 for your use case?

It sounds like an impressive list of features you've developed. My main reaction is that I am trying to minimize my maintenance burden from this package and others so I can focus on the development of [ITensorNetworks.jl](https://github.com/mtfishman/ITensorNetworks.jl), which has our next generation MPS, TTN, and PEPS/general TN solver and evolution code, using a graph-inspired interface for tensor networks. My goal there is to take lessons from this package like the `CelledVector` design and design a general infinite tensor network type, with infinite MPS/VUMPS just as a special case.

So with that in mind, let me go through your list:

- SU(3) and SU(4) irrep using color conservation (nothing non-Abelian)

Can this just be shared as an example code for others interested in using that? It doesn't sound like it requires modifications to VUMPS/InfiniteMPS itself.

- some small fixes in general

General small fixes would definitely be appreciated (maybe split up into PRs with fixes that are thematically similar).

- implementation of a few convenient functions for data analysis (QN-resolved analysis of ent spectrum, transfer matrix spectrum + some correlation functions)

I was hoping to provide this kind of functionality as examples at first, and think of this package as more bare-bones. For example, my goal was to provide an InfiniteMPS type and VUMPS but not much else, but provide examples and a nice interface so people can do that kind of thing (measurements and analysis of the infinite MPS) themselves. Also, I was hoping that by making it easy to slice an infinite MPS into a finite MPS then users can leverage finite MPS tools from ITensors.jl to perform many useful operations for analyzing the state. The developer burden of maintaining and improving the MPS/MPO code in ITensors.jl was immense. Things are more stable in that package now, but since I am trying to push the development of `ITensorNetworks.jl` as a replacement for the existing MPS/MPO code in ITensors.jl (and ultimately a replacement for this package), and also maintain the other packages in the ITensors.jl ecosystem, I don't want to double that effort by providing and maintaining too many features in this package. So I think it is best to share specialized tools like that in examples unless there is something that crosses the threshold of being very general and that has multiple use cases. If anything, my goal would be to try to _compress_ the code of this package where possible, as I discuss below.

- a working iDMRG for InfiniteMPOMatrix

That sounds great. I would appreciate a PR for that. If the code is simple and it is well tested it would make sense to include in this package. However, I think we should either choose `InfiniteMPO` or `InfiniteMPOMatrix` and stick with that, or make the codes only very minimally different, in order to minimize maintenance burden.

- a working InfiniteMPO constructed from an InfiniteMPOMatrix

That sounds useful. I think it is unfortunate how many different MPO formats there are in this package, and makes maintaining the package challenging. I think it would be best if we stuck with just one external format if possible, and a minimal number of internal formats, and try to design it so that we just have one version of VUMPS/iDMRG to maintain, optimize, etc. For example, perhaps you can appreciate the challenges involved given the number of performance PRs I had to make: #56, #64, #65, #67. I think this would have been largely unnecessary if we had one VUMPS implementation with a single infinite MPO format. Do you think at this point we can just go with one format? Is there a reason to keep around both `InfiniteMPO` and `InfiniteMPOMatrix`? Additionally, I think we should drop support for `InfiniteSum{MPO}`/`InfiniteSum{ITensor}` since I would hope those become largely unnecessary with good support for a single infinite MPO of some form.

- a working iDMRG for InfiniteMPO

Same comments as above.

- compression of iMPOs using InfiniteMPOMatrix and the approach by Parker, Cao et Zaletel in [Phys.Rev.B 102](https://doi.org/10.1103/PhysRevB.102.035147) . It is not yet fully benchmarked so that should clearly be _experimental_ , especially as it is the simplest, non-fail proof version.

I think @JanReimers has a version of that algorithm working in [ITensorMPOCompression.jl](https://github.com/JanReimers/ITensorMPOCompression.jl). Can you join forces on that? @JanReimers, is the infinite MPO version working?

- I also plan to do this week some modifications on the subspace expansion as I cannot afford the default version in my new project (the one that required the compression of iMPOs)

That sounds great. Improved subspace expansion tools would be very appreciated, hopefully they can also be shared with the solvers in `ITensorTDVP.jl`/`ITensorNetworks.jl` since those would also benefit from better subspace expansion methods. @b-kloss and @emstoudenmire have worked on some subspace expansion methods in https://github.com/ITensor/ITensorTDVP.jl/pulls so I'm curous how it relates and if it has some analog for finite MPS.

mtfishman:
Also, I think the issue was from a bug in the JuliaFormatter package which is now fixed, hopefully running the formatter again will fix the CI issue.

nbaldelli:
Hi @mtfishman, yes this looks like a cleaner solution of what I was trying in my fork, I will try it and close the issue. Thanks @LHerviou !

JanReimers:
On Thu, Feb 23, 2023 at 10:11 PM Matt Fishman ***@***.***>
wrote:

>
>    - compression of iMPOs using InfiniteMPOMatrix and the approach by
>    Parker, Cao et Zaletel in Phys.Rev.B 102
>    <https://doi.org/10.1103/PhysRevB.102.035147> . It is not yet fully
>    benchmarked so that should clearly be *experimental* , especially as
>    it is the simplest, non-fail proof version.
>
> I think @JanReimers <https://github.com/JanReimers> has a version of that
> algorithm working in ITensorMPOCompression.jl
> <https://github.com/JanReimers/ITensorMPOCompression.jl>. Can you join
> forces on that? @JanReimers <https://github.com/JanReimers>, is the
> infinite MPO version working?
>
Yes it is working for InfiniteMPO with dense storage and for some cases
(Trans. Ising) with block sparse storage.  The caveats are:

   1. It assumes all the new QR/RQ code is in ITensors.  But I think those
   changes are still in the PR queue.
   2. For block sparse storage the combiners used for SVD/QR/RQ will in
   general re-order and combine the QN spaces, breaking regular form.  So
   right now I am working on gauge transforming back to regular form after the
   decomp operations.  For simple systems like transverse Ising this problem
   does not arise. On the other hand, for a Hubbard model the spaces are much
   more complicated so I am using that as a benchmark for success.

In any case I am happy to coordinate with @LHerviou
<https://github.com/LHerviou>! @nbaldelli <https://github.com/nbaldelli> .
Perhaps their work has eclipsed what I have been doing, in which case I can
possibly contribute with documentation work.

Regards
Jan

mingruyang:
`function initstate(n)
    if n%2==0
        if (n/2)%2==1 
            return "Dn"
        else 
            return "Up"
        end
    else
        if (n+1)/2<=19
            if ((n+1)/2)%2==1
                return "Up"
            else
                return "Dn"
            end
        else 
            return "Emp"
        end
    end
end

N = 62
s = infsiteinds(n->isodd(n) ? "tJ" : "S=1/2", N; conserve_qns, initstate)
ψ = InfMPS(s, initstate)`

I got
`AssertionError: Flux not invariant under one unit cell translation, not implemented`

mtfishman:
@LHerviou @JanReimers I've moved discussions about other topics (InfiniteMPO generation and compression, iDMRG, subspace expansion, etc.) to #73. Let's keep this PR specific to the improvements to `infsiteinds`.

nbaldelli:
Hi @LHerviou, any update on this? Otherwise I could try to merge the same fix from my fork.

LHerviou:
Sorry, somehow everything went into my spams.

I will submit my fix today or tomorrow.
Sorry for the delay

LHerviou:
Here it is. It passed all the tests on my branch.

A few things:
- I improved how it deals with Z_p symmetries. This results sometimes in weird looking fillings, though accurate.
- It optimizes the multiplying factor for each QN separately. In the example above, that means (Nf, -19) or (Nf, 43) for the fermionic charge ( there are 19 occupied sites, and 43 free unoccupied), but keeping (Sz, -1),  (Sz, 0), (Sz, 1) for the spin QN (the total Sz is fixed).

mingruyang:
There is still problem with this fix.
If then I call
```
ψ = vumps_subspace_expansion(H, ψ; outer_iters, subspace_expansion_kwargs, vumps_kwargs)
```
I will get an error message:
```
Impossible to do a subspace expansion, probably due to conservation constraints
```

LHerviou:
If I am not mistaken, this is independent of the fix, but a limitation
Could you just give me the tight binding form of the Hamiltonian?
Also, which form of H are you using?

Proposed explanation: due to the two U(1) symmetries, blocking occurs and the two-site subspace expansion which is implemented by default cannot expand the bond dimension/the state.
=> typical example, pair hopping c^\dagger_j c_{j+1} c_{j+2} c^\dagger_{j+3} + h.c.

mingruyang:
```
function ITensorInfiniteMPS.unit_cell_terms(::Model"mymodel"; t, Jc, Js, Jk, Jcs, N)
  ampo = OpSum()

    for j in 1:2:N
        ampo += -t,"Cdagup",j,"Cup",j+2;
        ampo += -t,"Cdagdn",j,"Cdn",j+2;
        ampo += -t,"Cdagup",j+2,"Cup",j;
        ampo += -t,"Cdagdn",j+2,"Cdn",j;
        ampo += 0.5*Jc,"S+",j,"S-",j+2;
        ampo += 0.5*Jc,"S-",j,"S+",j+2;
        ampo +=     Jc,"Sz",j,"Sz",j+2;
        ampo += -0.25*Jc,"Ntot",j,"Ntot",j+2;
    end
    for j in 2:2:N
        ampo += 0.5*Js,"S+",j,"S-",j+2;
        ampo += 0.5*Js,"S-",j,"S+",j+2;
        ampo +=     Js,"Sz",j,"Sz",j+2;
    end
    for j in 1:2:N
        ampo += 0.5*Jk,"S+",j,"S-",j+1;
        ampo += 0.5*Jk,"S-",j,"S+",j+1;
        ampo +=     Jk,"Sz",j,"Sz",j+1;
    end
    for j in 1:2:N
        ampo += 0.5*Jcs,"S+",j,"S-",j+3;
        ampo += 0.5*Jcs,"S-",j,"S+",j+3;
        ampo +=     Jcs,"Sz",j,"Sz",j+3;
    end
    for j in 2:2:N
        ampo += 0.5*Jcs,"S+",j,"S-",j+1;
        ampo += 0.5*Jcs,"S-",j,"S+",j+1;
        ampo +=     Jcs,"Sz",j,"Sz",j+1;
    end

  return ampo
end
model = Model("mymodel")

H = InfiniteSum{MPO}(model, s; N)
```

mingruyang:
Probably because the hopping only happens between next-nearest neighbor sites. But at least it needs to give me a more clear and informative error message than that.

LHerviou:
It is exactly that, normally. If you have an idea for a better error message, feel free to suggest one.

LHerviou:
To expand a bit:
Actually, when I run your code on my version, I do get a subspace expansion for the sites before 19, because the nearest neighbours term J_K can act.
Because of the choice of initialization, the nearest neighbor Hamiltonian acts trivially on the second half of the chain (i.e., it does nothing as the fermionic sites are empty). Then we get stuck because the only non-trivial terms are nearest neighbours.

mingruyang:
For example, as you have mentioned, the error is from that the default two-site subspace expansion does not work in this case, so probably the error message could be "Impossible for the default two-site scheme to do a subspace expansion  fails due to the form of the QN conserved Hamiltonian" or pointing our at which line and inside which function the subspace expansion fails. The current error message is difficult for people who are not developer of this repository to trace back where the error happens.

src/infinitecanonicalmps.jl
```diff
@@ -48,11 +60,82 @@ function shift_flux(i::Index, flux_density::QN)
   return ITensors.setspace(i, shift_flux(space(i), flux_density))
 end
 
+function multiply_flux(qn::ITensors.QNVal, flux_factor::Int64)
```
mtfishman, line 63: I would prefer the name `scale_flux`. Also maybe allow more general `flux_factor` inputs with `::Integer`.

src/infinitecanonicalmps.jl
```diff
@@ -48,11 +60,82 @@ function shift_flux(i::Index, flux_density::QN)
...
+  elseif qn.modulus == 1 || qn.modulus == -1 #U(1) symmetry, should remain a U(1) symmetry
+    return (qn.name, qn.val * flux_factor, qn.modulus) #qn * flux_factor
+  else #To get the same algebra, multiplying the modulus by flux_factor works
+    return (qn.name, qn.val * flux_factor, qn.modulus * flux_factor)
+  end
+end
+
+function multiply_flux(qn::QN, flux_factor::Int64)
+  flux_factor == 1 && return qn
+  return QN(multiply_flux.(filter(x -> x.modulus != 0, qn.data), flux_factor)...)
```
mtfishman, line 75: Can't the logic of checking `flux_factor == 1` and `x.modulus != 0` be handled by `multiply_flux` (to be renamed `scale_flux`)?

Also, why would one of the QNs have a modulus of 0? It seems like that should be caught somewhere else and not filtered/checked here.

src/infinitecanonicalmps.jl
```diff
@@ -48,11 +60,82 @@ function shift_flux(i::Index, flux_density::QN)
...
+end
+
+function multiply_flux(qn::QN, flux_factor::Dict{ITensors.SmallString,Int64})
+  return QN(
+    [multiply_flux(q, flux_factor[q.name]) for q in filter(x -> x.modulus != 0, qn.data)]...
```
mtfishman, line 80: Again, I'm confused about the check/filter for `x.modulus != 0`.

src/infinitecanonicalmps.jl
```diff
@@ -48,11 +60,82 @@ function shift_flux(i::Index, flux_density::QN)
...
+  if iszero(flux)
+    return s
+  end
+  #For simplicity, we are not introducing a factor per QN
+  n = length(s)
+  multipliers = Dict{ITensors.SmallString,Int64}() #multipliers is assigned using the names
+  for qn in flux.data
+    if qn.modulus == 0
+      continue
+    end
```
mtfishman, line 134: It looks like if we check for `qn.modulus == 0` we don't need the checks for that in the functions above (except for maybe the core function `scale_flux(qn::ITensors.QNVal, flux_factor::Integer)`). However, I would be curious to hear where that case comes up.

mtfishman:
@mingruyang could you start a new issue for problems you come across with subspace expansion (with a minimal runnable code example)?

LHerviou:
The simplest test is probably to slightly modify our starting state for Ising or XXZ and check that it knows how to build the correct starting MPS. For example, starting with 3 fermions and 1 hole in XXZ with nsites = 4.

As a side comment, the implementation for Z_n conserved quantities could be improved easily, if you think it is relevant, by instead finding y such that y * nsites = flux [n], if such solution exists.

mtfishman:
That would be a nice modification to the tests.

That sounds like a good idea to improve the handling of Zn symmetries!

LHerviou:
I will try to do this this weekend then.

LHerviou:
Normally this works perfectly, and was running on my side.

It relies on the Chinese remainder theorem for Zp symmetries, i.e. noting N the unit cell, we tried to find y such that 
y = 0 mod N and y = flux mod p. Then we just need to find x such that y = N x.

This system admits a solution iff flux = 0 mod( gcd(p, N)), which is not necessarily true. We can find a global multiplier b such that it is true.
The action of the multiplier is flux -> b flux, p -> b p.  b = N is a trivial solution. That was more or less what the previous code was doing.
Now I am a bit more subtle: I try to find the smallest b such that:
b flux = 0 mod gcd(N, p b).
It changes nothing but is more satisfactory and natural.

Case when things changed: N = 3, initstate = +--, flux = 1 mod 2.
Previous version: 1 is not divisible by 3, so we mutiply everything by 3. Gives QN ( 4 mod 6, 1 mod 6)
Current version: gcd(2, 3) = 1, so a solution actually exists: 3 x 1 = 1 mod 2 => work with QN (0 mod 2, 1 mod 2)

test/test_unitcell_qns.jl
```diff
@@ -0,0 +1,47 @@
...
+    else
+      return 1
+    end
+  end
+
+  for N in 1:6
+    s = infsiteinds("S=1/2", N; conserve_szparity=true, initstate)
+    ψ = InfMPS(s, initstate)
+    @show N
+    @show s
```
mtfishman, line 45: Is there a test we can add here, or are we just checking that the constructor doesn't error? Maybe we can test the flux of a unit cell of the MPS?

LHerviou:
Well, the construction of \psi already guarantees that the flux is 0. But I added the @test in any case.

mtfishman:
Thanks @LHerviou! Is this good to merge from your end once tests pass?

LHerviou:
Yes !

PR #70, opened 2023-02-27
Fix sweeps in approx mps

ryanlevy:
Small typo in `infinitemps_approx.jl`, this should let the sweeps be variable


mtfishman:
Thanks!

Issue #71, opened 2023-02-27
Slicing an infinite MPS to get a finite one 


nbaldelli:
Following the discussion in #69 I tried some ways to get a finite MPS from Infinite ones. 

Directly slicing via `psi_finite = psi[start:stop]` creates problems with the dangling Link indices at the beginning and the end of the resulting MPS when using functions designed for finite MPS, i.e. `correlation_matrix`. 

The solution I would implement would be creating a finite MPS by attaching two "Dummy" indices at the beginning and the end of the chain via delta tensors to close the dangling links, after having put the finite slice in a proper canonical form. One would only need to take the precaution not to do any operation on the dummy sites.

I tried it and it works when computing the correlation function, both using the function in ITensorCorrelators (ITensors `correlation_matrix` automatically applies operators to the first site of the MPS) and by directly applying the operators. 

I plan on making a PR with an example file with a function `toMPS` and some tests but open this issue in case something else comes up.

nbaldelli:
The issue with dangling indices in `correlation_matrix` is solved by https://github.com/ITensor/ITensors.jl/pull/1081. The Infinite MPS -> Finite MPS issue is solved by #72

PR #72, opened 2023-02-27
Added example file with Infinite MPS to finite MPS function

nbaldelli:
Related to #71.

I mainly added a function `toMPS` that 

1. takes a slice of an `InfiniteCanonicalMPS`
2. puts it in a canonical form with the change of gauge on the last link
3. adds two copy tensors at the beginning and end of the chain to add two "Dummy" sites to contract the dangling Link indices of the slice. 
4. It finally returns a finite MPS.

I added two functions computing the correlation matrix (by explicitly applying the operators) for both the infinite and infinite chain to check for consistency. I did not add the comparison with the correlation function from `ITensorCorrelators` as It would add a dependency.


mtfishman:
Thanks @nbaldelli, I like the idea of adding "dummy tensors" at the boundaries so it looks more like a finite MPS without dangling indices. Here is a simpler implementation:
```julia
function finite_mps(ψ::InfiniteCanonicalMPS, range::AbstractRange)
  @assert isone(step(range))
  ψ_finite = ψ.AL[range]
  ψ_finite[last(range)] *= ψ.C[last(range)]
  l0 = linkind(ψ.AL, first(range) - 1 => first(range))
  lN = linkind(ψ.AR, last(range) => last(range) + 1)
  ψ_finite = MPS([δ(l0, dag(l0)'); [ψ_finiteᵢ for ψ_finiteᵢ in ψ_finite]; δ(dag(lN), lN')])
  set_ortho_lims!(ψ_finite, (last(range) + 1):(last(range) + 1))
  return ψ_finite
end
```
This seems worth adding to the package. Could you make a PR for that, with a simple test? Then we don't need to implement this as a separate example, but just add the use of `finite_mps` to an existing example.

I noticed that when I do:
```julia
range = 1:6
correlation_matrix(finite_mps(ψ, range), "Sz", "Sz"; sites=(range .+ 1))
```
`correlation_matrix` throws an error about trying to make an `"Sz"` operator on site 1 (a site with the link index). It shouldn't have to do that in principle, it is a simple fix to make sure it skips that site (currently it just uses site 1 by default to analyze the Hermiticity of the operator, but we can just use `first(sites)`).

mtfishman:
@nbaldelli I fixed the issue with `correlation_matrix` in https://github.com/ITensor/ITensors.jl/pull/1081 so now the code ```julia
range = 1:6
correlation_matrix(finite_mps(ψ, range), "Sz", "Sz"; sites=(range .+ 1))
```
should work.

**Comment by nbaldelli** [comment]
Hi Matt, just tested the code with your formulation of `finite_mps` and the new fix in https://github.com/ITensor/ITensors.jl/pull/1081 and it works, I pushed the new changes and moved the test to the Ising model example file.

File: examples/vumps/vumps_ising.jl [hunk 1]
```diff
@@ -102,3 +102,72 @@ Sz2_infinite = expect(ψ.AL[2] * ψ.C[2], "Sz")
...
+function finite_mps(ψ::InfiniteCanonicalMPS, range::AbstractRange)
+  @assert isone(step(range))
+  ψ_finite = ψ.AL[range]
+  ψ_finite[last(range)] *= ψ.C[last(range)]
+  l0 = linkind(ψ.AL, first(range) - 1 => first(range))
+  lN = linkind(ψ.AR, last(range) => last(range) + 1)
+  ψ_finite = MPS([δ(l0, dag(l0)'); [ψ_finiteᵢ for ψ_finiteᵢ in ψ_finite]; δ(dag(lN), lN')])
+  set_ortho_lims!(ψ_finite, (last(range) + 1):(last(range) + 1))
+  return ψ_finite
+end
```
mtfishman, line 118: Let's move this into the package (say to `src/infinitecanonicalmps.jl`) and export it (add it to the export list in `src/ITensorInfiniteMPS.jl`).

examples/vumps/vumps_ising.jl
```diff
@@ -102,3 +102,72 @@ Sz2_infinite = expect(ψ.AL[2] * ψ.C[2], "Sz")
...
+          h = op(op1, siteinds(ψ)[i]) * op(op2, siteinds(ψ)[j])
+          ϕ = ψ[i]
+          for k in (i+1):(j)
+              ϕ *= ψ[k]
+          end
+          C[i-start+1,j-start+1] = (noprime(ϕ * h) * dag(ϕ))[]
+      end
+  end
+  return C + C'
+end
```
mtfishman, line 157: I think we can just remove this and show how to use `ITensors.correlation_matrix`.

As a test we could just compare the result to a single correlator computed with an MPO.

examples/vumps/vumps_hubbard_extended.jl
```diff
@@ -1,5 +1,6 @@
 using ITensors
 using ITensorInfiniteMPS
+using Revise
```
mtfishman, line 3: We should remove this, you can load it by default on your own system (https://timholy.github.io/Revise.jl/stable/config/#Using-Revise-by-default-1).

examples/vumps/vumps_hubbard_extended.jl
```diff
@@ -120,19 +121,24 @@ Sz_finite = ITensors.expect(ψfinite, "Sz")[nfinite:(nfinite + 1)]
 
 energy_exact = reference(model, Observable("energy"); U=model_params.U / model_params.t)
 
+corr_infinite = ITensors.correlation_matrix(finite_mps(ψ, 1:10), "Nup", "Ndn")
```
mtfishman, line 124: Should be able to change `ITensors.correlation_matrix` to `correlation_matrix` since it is exported from `ITensors.jl`.

mtfishman: Also I think you need to add the keyword argument `sites=2:11` to skip the first "site" which has a link index.

examples/vumps/vumps_ising.jl
```diff
@@ -102,3 +102,19 @@ Sz2_infinite = expect(ψ.AL[2] * ψ.C[2], "Sz")
 
 @show Sz1_finite, Sz2_finite
 @show Sz1_infinite, Sz2_infinite
+
+start = 1; stop = 6 #where to slice the infiniteMPS
+psi = finite_mps(ψ, start:stop)
+
+A = correlation_matrix_gates(psi, "Sz", "Sz", start+1, stop+1)
```
mtfishman, line 109: Should be dropped now.

examples/vumps/vumps_ising.jl
```diff
@@ -102,3 +102,19 @@ Sz2_infinite = expect(ψ.AL[2] * ψ.C[2], "Sz")
...
+B = ITensors.correlation_matrix(psi, "Sz", "Sz", sites=(start+1):(stop+1))
```
mtfishman, line 110: Change to just `correlation_matrix(...)`.

mtfishman:
Thanks @nbaldelli. Could you run the formatter to format the code? You can use:
```julia
julia> using JuliaFormatter, ITensorInfiniteMPS; format(pkgdir(ITensorInfiniteMPS))
```

src/infinitecanonicalmps.jl
```diff
@@ -148,6 +148,17 @@ function InfMPS(s::CelledVector, f::Function)
...
+function finite_mps(ψ::InfiniteCanonicalMPS, range::AbstractRange)
+  @assert isone(step(range))
+  ψ_finite = ψ.AL[range]
+  ψ_finite[last(range)] *= ψ.C[last(range)]
+  l0 = linkind(ψ.AL, first(range) - 1 => first(range))
+  lN = linkind(ψ.AR, last(range) => last(range) + 1)
+  ψ_finite = MPS([δ(l0, dag(l0)'); [ψ_finiteᵢ for ψ_finiteᵢ in ψ_finite]; δ(dag(lN), lN')])
+  set_ortho_lims!(ψ_finite, (last(range) + 1):(last(range) + 1))
+  return ψ_finite
+end
```
mtfishman, line 160: Here is a slight improvement on this function:
```julia
function finite_mps(ψ::InfiniteCanonicalMPS, range::AbstractRange)
  @assert isone(step(range))
  N = length(range)
  ψ_finite = ψ.AL[range]
  ψ_finite[N] *= ψ.C[last(range)]
  l0 = linkind(ψ.AL, first(range) - 1 => first(range))
  l̃0 = sim(l0)
  lN = linkind(ψ.AR, last(range) => last(range) + 1)
  l̃N = sim(lN)
  δl0 = δ(dag(l̃0), l0)
  δlN = δ(dag(l̃N), lN)
  ψ_finite[1] *= δl0
  ψ_finite[N] *= dag(δlN)
  ψ_finite = MPS([dag(δl0); [ψ_finiteᵢ for ψ_finiteᵢ in ψ_finite]; δlN])
  set_ortho_lims!(ψ_finite, (N + 1):(N + 1))
  return ψ_finite
end
```
I think previously it was slightly incorrect with the indexing if the `range` input didn't start at 1. Also, this makes it so that the first and last "site" indices of the finite MPS exactly match the dangling link indices of the infinite MPS slice (before they were primes of those indices).

nbaldelli:
Cleaning up the code, in the end I moved the example to `vumps_hubbard_extended.jl` where there is already a comparison with the finite DMRG results.

nbaldelli:
@mtfishman I think everything is fine now

mtfishman:
It looks like you still need to add `finite_mps` to the export list in `src/ITensorInfiniteMPS.jl`.

nbaldelli:
So the error in the test caused by the fact that the function `finite_mps` is not recognized unless I call it via `ITensorInfiniteMPS.finite_mps`, should I add something to the exports page? `infinitecanonicalmps.jl` is already exported

mtfishman:
Great, thanks @nbaldelli!

PR #76, opened 2023-04-02
InfiniteMPOMatrix -> InfiniteMPO conversion

JanReimers:
Started with Lioc Herviou's code. Uses directsum() to build the InfiniteMPO tensors. Supports dense and block-sparse under the hood (no if hasqns() code blocks).  Tested with Heisenberg/Hubbard models with 1-5 NN interactions and Lioc's FQHE model with Ncell=6 and Ly=3.  
This is not the most efficient way to build InfiniteMPO, the output has extra columns and rows with I's and all zeros that should subsequently  be removed with block repsecting orthogonalization and or compression.  We should be able to build a more efficient  InfiniteMPO by using a block respecting directsum as described in eq. E3 in the Parker paper.  But of users are going to compress anyway, maybe this is not useful.

### DIFF
--- a/src/infinitempomatrix.jl
+++ b/src/infinitempomatrix.jl
@@ -60,53 +60,53 @@ end
 #  This code should be 100% dense/blocks-sparse agnostic.
 #
 function matrixITensorToITensor(Hm::Matrix{ITensor})::ITensor
-  indexT=typeof(inds(Hm[1,1])[1])
-  T=eltype(Hm[1,1])
+  indexT = typeof(inds(Hm[1, 1])[1])
+  T = eltype(Hm[1, 1])
   lx, ly = size(Hm)
-  @assert lx==ly
+  @assert lx == ly
   #
   #  Extract the sub diagonal
   #
-  Ms=map(n->Hm[lx-n+1,ly-n],1:lx-1) #Get the sub diagonal into an array.
-  N=length(Ms)
-  @assert N==lx-1
+  Ms = map(n -> Hm[lx - n + 1, ly - n], 1:(lx - 1)) #Get the sub diagonal into an array.
+  N = length(Ms)
+  @assert N == lx - 1
   #
   #  We need a dummy link index, but it has to have the right direction.  
   #  Ms[1] should only have one link index which should be pointing to the right.
   #
-  @assert length(inds(Ms[1],tags="Link"))==1
-  ir,=inds(Ms[1],tags="Link")
-  il0=Index(ITensors.trivial_space(ir);dir=dir(dag(ir)),tags="Link,l=0") #Dw=1 left dummy index.
+  @assert length(inds(Ms[1]; tags="Link")) == 1
+  ir, = inds(Ms[1]; tags="Link")
+  il0 = Index(ITensors.trivial_space(ir); dir=dir(dag(ir)), tags="Link,l=0") #Dw=1 left dummy index.
   #
   # Convert edge Ms to order 4 ITensors using the dummy index.
   #
-  Ms[1]*=onehot(T, il0 => 1)
-  Ms[N]*=onehot(T, dag(il0) => 1) #We don't need distinct index IDs for this.  directsum makes all new index anyway.
+  Ms[1] *= onehot(T, il0 => 1)
+  Ms[N] *= onehot(T, dag(il0) => 1) #We don't need distinct index IDs for this.  directsum makes all new index anyway.
   #
   #  Create list of all left and right link indices on the Ms.  In order to support dense storage we 
   #  can't use QN directions to decide left/right.  Instead we look for common tags among the neighbouring Ms.
   #  Right now the code assumes plev=0 for all Ms.
   #
-  is=NTuple{2,indexT}[]
-  ir=il0 #set up recursion below.  ir is from the previous M, should have the same tags as il on the next M.
+  is = NTuple{2,indexT}[]
+  ir = il0 #set up recursion below.  ir is from the previous M, should have the same tags as il on the next M.
   for n in 1:N
-    il,=inds(Ms[n],tags=tags(ir)) #Find the left index
-    ir=noncommonind(Ms[n],il,tags="Link") #Find the new right index by elimination
-    push!(is,(il,ir))    
-  end 
+    il, = inds(Ms[n]; tags=tags(ir)) #Find the left index
+    ir = noncommonind(Ms[n], il; tags="Link") #Find the new right index by elimination
+    push!(is, (il, ir))
+  end
   #
   # Now direct sum everything non-empty in Hm, including those I ops in the corners.
   # Order is critical.   
   # Details of the Link index tags like "l=n" do not matter from here on.
   #
-  H=Hm[1,1]*onehot(T, il0' => 1)*onehot(T, dag(il0) => 1) #Bootstrap with the I op in the top left of Hm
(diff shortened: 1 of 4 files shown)


src/infinitempomatrix.jl
```diff
@@ -43,4 +44,91 @@ function InfiniteMPOMatrix(data::Vector{Matrix{ITensor}}, translator::Function)
...
+#    :         :                     :           :      :
+#    0         0          ...  l=2-->M-->l=1     0      0
+#    0         0          ...        0        l=1-->M   I
+#
+#  We need to capture the corner I's and the sub-diagonal Ms, and paste them together into an ITensor.
+#  This is facilutated by making all elements of Hm into order(4) tensors by adding dummy Dw=1 indices.
+#  We ITensors.directsum() to join all the blocks into the final ITensor.
+#  This code should be 100% dense/blocks-sparse agnostic.
+#
+function matrixITensorToITensor(Hm::Matrix{ITensor})::ITensor
```
mtfishman, line 62: I would call this `cat_to_itensor(Hm::Matrix{ITensor})`, in analogy to Julia's block matrix concatenation interface:
```julia
julia> x = [[randn(2, 2)] [randn(2, 2)]; [randn(2, 2)] [randn(2, 2)]]
2×2 Matrix{Matrix{Float64}}:
 [-0.53919 -0.0631672; 1.70176 1.02657]  [-0.329444 1.71233; 1.68521 -0.113242]
 [0.0322956 0.240616; 2.24649 1.24855]   [-0.445636 1.02609; 1.72244 -0.10569]

julia> hvcat((2, 2), x...)
4×4 Matrix{Float64}:
 -0.53919   -0.0631672   0.0322956   0.240616
  1.70176    1.02657     2.24649     1.24855
 -0.329444   1.71233    -0.445636    1.02609
  1.68521   -0.113242    1.72244    -0.10569

julia> [x[1, 1] x[1, 2]; x[2, 1] x[2, 2]]
4×4 Matrix{Float64}:
 -0.53919    -0.0631672  -0.329444   1.71233
  1.70176     1.02657     1.68521   -0.113242
  0.0322956   0.240616   -0.445636   1.02609
  2.24649     1.24855     1.72244   -0.10569
```
(we can see this function as an ITensor version of that kind of concatenation).

src/infinitempomatrix.jl
```diff
@@ -43,4 +44,91 @@ function InfiniteMPOMatrix(data::Vector{Matrix{ITensor}}, translator::Function)
...
+  #  Extract the sub diagonal
+  #
+  Ms = map(n -> Hm[lx - n + 1, ly - n], 1:(lx - 1)) #Get the sub diagonal into an array.
+  N = length(Ms)
+  @assert N == lx - 1
+  #
+  #  We need a dummy link index, but it has to have the right direction.  
+  #  Ms[1] should only have one link index which should be pointing to the right.
+  #
+  @assert length(inds(Ms[1]; tags="Link")) == 1
```
mtfishman, line 77: Please avoid using `tags` in this way for filtering or extracting indices (that goes for the rest of the PR as well).

src/infinitempomatrix.jl
JanReimers: Done

src/infinitempomatrix.jl
JanReimers: Assuming you are talking about the semicolon: Done

mtfishman: No, I meant a deeper change, that instead of finding a certain Index using the tags of the Index, we should find indices by searching which indices of the tensor are shared with neighboring tensors in the network. So for example in a line like this:
```julia
il, = inds(Ms[n], tags=tags(ir)) #Find the left index
```
we should have instead already determined which are the left and right indices of the MPO tensors by analyzing the MPO networks and extracting the `linkinds` and then passing those indices into this function (maybe there is a unique way to determine which indices are getting direct summed over/are the link indices instead of the site indices without passing indices into this function and without using the tags of the indices, but I doubt it).

Basically, we should take as an axiom or design "law" of ITensor code that we should not use tags to find indices, filter indices, decide which indices to prime, etc. and instead use tensor network analysis tools or index set operation like `linkinds`, `commoninds`, `uniqueinds`, etc. (our versions of what Julia would call Set operations: https://docs.julialang.org/en/v1/base/collections/index.html#Set-Like-Collections). This basically means not using the `tags` keyword argument for filtering indices _at all_. I have found that designing code in this way is much more robust, leads to simpler code in the end, the code is less error prone, etc.

Tags should just be treated as being used mostly for printing, or as ways to avoid having indices contract with each other (which is the purpose of the call tags of the InfiniteMPS).

mtfishman: For the record, for passing keyword arguments to functions, I prefer the semicolon syntax. This is also consistent with the "Blue" style guide that we follow in our code and which is enforced by the auto formatter: https://github.com/invenia/BlueStyle#keyword-arguments

JanReimers: I think this would require mods to the generation of InfiniteMPOMatrix:
Hm = InfiniteMPOMatrix(model, s; model_params...);
Ms = map(n -> Hm[1][lx - n + 1, ly - n], 1:(lx - 1))
@show inds(Ms[1],tags="Link")
@show inds(Ms[2],tags="Link")
@show inds(Ms[3],tags="Link")

inds(Ms[1], tags = "Link") = ((dim=5|id=553|"Link,l=1") 
inds(Ms[2], tags = "Link") = ((dim=5|id=154|"Link,l=1")  (dim=7|id=906|"Link,l=2") 
inds(Ms[3], tags = "Link") = ((dim=7|id=582|"Link,l=2")  (dim=11|id=172|"Link,l=3") 

The IDs don't line up, only the tags.

mtfishman: The `InfiniteMPOMatrix` was constructed from sets of MPOs which had common indices, so there must be common indices across the sites of the `InfiniteMPOMatrix`. Specifically, if you did a matrix multiplication of the `Matrix{ITensor}`s on site `1` and `2`, i.e. `Hm[1] * Hm[2]`, shouldn't the tensors making up the matrices `Hm[1]` and `Hm[2]` share common link indices? Those are the common indices I'm taking about. I don't think we need to (or should) assume that within a certain `Matrix{ITensor}`, say `Hm[1]`, that the tensors share common link indices (though they should share common site indices). We should pass information into the function about which indices are shared between `Hm[1]` and `Hm[2]`, for example, and then direct sum over those indices.

JanReimers: OK I get it now ... thanks for your patience.

JanReimers: Done

mtfishman:
Looks great, thanks @JanReimers! Glad to see it ended up so simple.

I recall that you said something about having to make use of some slicing functionality from your package https://github.com/JanReimers/ITensorMPOCompression.jl to implement this function. I don't see that here, is that no longer necessary?

JanReimers:
Yes, directsum made slicing unnecessary.

mtfishman:
> Yes, directsum made slicing unnecessary.

Great, thanks for clarifying. I imagine that slicing could be faster, at least the way `directsum` is implemented right now, so if we see any performance issues we could keep that in mind. `directsum` itself could be sped up with better slicing support (the implementation right now is simple but not designed for efficiency) so that is also something to keep in mind as a potential improvement.

JanReimers:
Yes my anecdotal example for that is that early versions of the compression module did all the slicing at the ITensors level.  When I moved it down to the NDTensors level (with subtensors.jl) the performance improvement was stunning.

PR #77, opened 2023-06-15
Improving infinite mpo matrix

LHerviou:
As described, some redefinitions and improvements on the form of InfiniteMPOMatrix, and modifications that follow.
The core idea:
- Arbitrary filled InfiniteMPOMatrix are allowed
- Instead of waiting for nrange(mpo) to evaluate the terms in the local MPO, we do not delay evaluation. T
- Lines and columns in an InfiniteMPOMatrix have well defined indices, shared for all elements.

All of that normally significantly simplifies the manipulation and the optimization of the InfiniteMPOMatrix and the corresponding InfiniteMPO
I did not have time to fully debug, so I will keep updating in the next few days, and confirm when one can merge.


src/models/fqhe13.jl
```diff
@@ -1,12 +1,47 @@
...
+      QN(("Nf", q - p), ("NfMom", (q - p) * pos)) => 1,
+    ]
+  end
+end
+
+# Forward all op definitions to Fermion
+function ITensors.op!(Op::ITensor, opname::OpName, ::SiteType"FermionK", s::Index...)
+  return ITensors.op!(Op, opname, SiteType("Fermion"), s...)
+end
+ITensors.has_fermion_string(::OpName"C", ::SiteType"FermionK") = true
```
mtfishman, line 16: Maybe we could use:
```julia
ITensors.has_fermion_string(opname::OpName, ::SiteType"FermionK") = has_fermion_string(opname, SiteType("Fermion"))
```
so it just uses the same definitions we already have in ITensors.jl.

LHerviou: Indeed, completely obvious and missed by me.

LHerviou:
All tests work on my branch without any measurable slowdown.

Probably a lot of simplification/optimization can be done. Up to you if you integrate it right now, or if we wait a bit til I can have an overview of possible changes.

src/models/models.jl
```diff
@@ -110,22 +116,113 @@ function InfiniteMPOMatrix(model::Model, s::CelledVector, translator::Function;
...
+          @assert iszero(temp_mat[1, 2])
+          @assert iszero(temp_mat[1, 3])
+          @assert iszero(temp_mat[2, 3])
+          @assert temp_mat[1, 1] == identity
+          @assert temp_mat[3, 3] == identity
+          Hmat[range_H + 1 - n, range_H - n] = temp_mat[2, 2]
+          Hmat[end, range_H - n] = temp_mat[3, 2]
+          Hmat[range_H + 1 - n, 1] = temp_mat[2, 1]
+          Hmat[range_H + 1 - n, range_H + 1 - n] *= ITensor(
+            T, filterinds(commoninds(temp_mat[2, 2], temp_mat[2, 1]); tags="Link")
```
mtfishman, line 144: Do we need the tag filter here? Best to try to avoid that.

src/infinitempomatrix.jl
```diff
@@ -44,111 +44,342 @@ function InfiniteMPOMatrix(data::Vector{Matrix{ITensor}}, translator::Function)
...
+    vector = fill(op("Zero", local_sit), 1, 3)
+  end
+  for (idx, proj) in enumerate([top, middle, bottom])
+    vector[idx] = proj * tensor
+  end
+  return vector
+end
+
+#Matrix multiplication for Matrix of ITensors
+function Base.:*(A::Matrix{ITensor}, B::Matrix{ITensor})
```
mtfishman, line 216: It looks like this "just works" without this definition if we overload `Base.zero(::ITensor)`:
```julia
julia> using ITensors

julia> Base.zero(T::ITensor) = false * T

julia> i, j, k = Index.((2, 2, 2))
((dim=2|id=221), (dim=2|id=645), (dim=2|id=568))

julia> A = randomITensor(i, j)
ITensor ord=2 (dim=2|id=221) (dim=2|id=645)
NDTensors.Dense{Float64, Vector{Float64}}

julia> B = randomITensor(j, k)
ITensor ord=2 (dim=2|id=645) (dim=2|id=568)
NDTensors.Dense{Float64, Vector{Float64}}

julia> [A A; A A] * [B; B]
2-element Vector{ITensor}:
 ITensor ord=2
Dim 1: (dim=2|id=221)
Dim 2: (dim=2|id=568)
NDTensors.Dense{Float64, Vector{Float64}}
 2×2
 -1.0347223964342365  -10.334190795304709
  0.6816120568533722    6.292266667641426
 ITensor ord=2
Dim 1: (dim=2|id=221)
Dim 2: (dim=2|id=568)
NDTensors.Dense{Float64, Vector{Float64}}
 2×2
 -1.0347223964342365  -10.334190795304709
  0.6816120568533722    6.292266667641426
```

src/infinitempomatrix.jl
```diff
@@ -44,111 +44,342 @@ function InfiniteMPOMatrix(data::Vector{Matrix{ITensor}}, translator::Function)
...
+        end
+      else
+        C[j] .+= A[k] * B[k, j]
+      end
+    end
+  end
+  return C
+end
+
+function scalar_product(A::Vector{ITensor}, B::Vector{ITensor})
```
mtfishman, line 300: With `Base.zero(::ITensor)` overloaded then this works:
```julia
julia> dot([A; A], [A; A])
22.549690454191992
```

src/infinitempomatrix.jl
LHerviou: Seems like a much simpler change.
Thanks for the advice!

mtfishman:
@LHerviou I'm fine waiting to merge this if you think there are simplifications or optimizations to make once you have the time, unless @JanReimers needs this for something and therefore it's useful to have merged.

JanReimers:
I have a branch with 

- InfiniteMPO compression outputting InfiniteCanonicalMPO
- VUMPS suppert for InfiniteMPO
- Simplification of VUMPS generic code
- VUMPS suppert for InfiniteCanonicalMPO (which may require some discussion)
I was volunteering to let Loic merge first and then I can deal with any resulting conflicts.

src/infinitempomatrix.jl
LHerviou: There are some definitions missing in NDTensors (contracting EmptyTensor with Diag tensor).
I had to include these definitions in the ITensors.jl file.

Else, everything seems to work with this change.

src/models/models.jl
LHerviou: I removed it. The lines were not strictly necessary anyway.

src/infinitempomatrix.jl
LHerviou: Ok, I am correcting what I said. 

Once I remove the definitions of + and * using only the functions you proposed, sometimes some weird bugs happen due to missing promote rules. 
A minimal example, I believe related but not directly equivalent:

```julia
using ITensors
using ITensorInfiniteMPS

is = Index([QN("N", 0)=>1, QN("N", 1)=>1])

a = ITensor(is, prime(is))
b = ITensor(prime(is, 2))
c = ITensor(is, prime(is), prime(is, 2))
cbis = ITensor(0, is, prime(is), prime(is, 2))

#Crash by default, with a StackOverflow of promote_rule
a*b + c

#No crash
a*b + cbis

#No crash
c + cbis
``` 

One of the reasons appear to be (as far as I can say) that a*b is not a correct tensor if we start from empty tensors.
By playing with promote_rule, I can manage to prevent the first crash, but then the previously working expression start failing.

mtfishman: Looks like a pretty subtle bug, your assessment is correct that the output type of `a * b` is a bit wrong there (the tensor type has the correct number of dimensions, but the storage type doesn't), and that seems to cause some ripple affect in `a * b + c` in type promotion. It seems like both `a * b` should be fixed, and also that kind of issue should be caught better in `a * b + c`.

Do you have an example of `Matrix{ITensor} * Matrix{ITensor}` failing with that kind of issue? That technically works for me but then issues arise if the result is used to add to other tensors because the output type is a bit incorrect:
```julia
julia> using ITensors
[ Info: Precompiling ITensors [9136182c-28ba-11e9-034c-db9fb085ebd5]

julia> i = Index([QN() => 2])
(dim=2|id=375) <Out>
 1: QN() => 2

julia> a = ITensor(i, i')
ITensor ord=2
(dim=2|id=375) <Out>
 1: QN() => 2
(dim=2|id=375)' <Out>
 1: QN() => 2
NDTensors.EmptyStorage{NDTensors.EmptyNumber, NDTensors.BlockSparse{NDTensors.EmptyNumber, Vector{NDTensors.EmptyNumber}, 2}}

julia> b = ITensor(i'')
ITensor ord=1
(dim=2|id=375)'' <Out>
 1: QN() => 2
NDTensors.EmptyStorage{NDTensors.EmptyNumber, NDTensors.BlockSparse{NDTensors.EmptyNumber, Vector{NDTensors.EmptyNumber}, 1}}

julia> x = [a a; a a] * [b b; b b]; # Work ok but has the wrong storage type

julia> x[1, 1] + ITensor(i, i', i'')
ERROR: StackOverflowError:
Stacktrace:
 [1] promote_result(#unused#::Type, #unused#::Type, #unused#::Type{NDTensors.EmptyStorage{NDTensors.EmptyNumber, NDTensors.BlockSparse{NDTensors.EmptyNumber, Vector{NDTensors.EmptyNumber}, 3}}}, #unused#::Type{NDTensors.EmptyStorage{NDTensors.EmptyNumber, NDTensors.BlockSparse{NDTensors.EmptyNumber, Vector{NDTensors.EmptyNumber}, 1}}})
   @ Base ./promotion.jl:321
 [2] promote_type
   @ ./promotion.jl:307 [inlined]
 [3] promote_result(#unused#::Type, #unused#::Type, #unused#::Type{NDTensors.EmptyStorage{NDTensors.EmptyNumber, NDTensors.BlockSparse{NDTensors.EmptyNumber, Vector{NDTensors.EmptyNumber}, 1}}}, #unused#::Type{NDTensors.EmptyStorage{NDTensors.EmptyNumber, NDTensors.BlockSparse{NDTensors.EmptyNumber, Vector{NDTensors.EmptyNumber}, 3}}})
   @ Base ./promotion.jl:321
```

mtfishman: @LHerviou I have a simple fix for this issue which makes sure the contraction code outputs the correct types, so then there is no issue in subsequent operations like addition. I'll make a PR to NDTensors/ITensors directly.

I can also add the definition `Base.zero(T::ITensor) = false * T` so then with those two fixes `Matrix{ITensor} * Matrix{ITensor}` will "just work".

LHerviou:
Sounds perfect. Does it allow also to remove the extension of datatype I
had to do?

Thanks a lot, Mat.

mtfishman:
I don't seem to need to overload `NDTensors.datatype` for `Matrix{ITensor} * Matrix{ITensor}` to work. Is it possible another operation needed that?

LHerviou:
I will check again tomorrow. It is possible it came from something else.

mtfishman:
@LHerviou the issues we discussed have been implemented in the latest version of ITensors.jl (https://github.com/ITensor/ITensors.jl/pull/1135).

LHerviou:
Apparently, there is still one definition missing with how you implemented zero.
```
julia> T = ITensor()
ITensor ord=0
NDTensors.EmptyStorage{NDTensors.EmptyNumber, NDTensors.Dense{NDTensors.EmptyNumber, Vector{NDTensors.EmptyNumber}}}

julia> zero(T)
ERROR: MethodError: no method matching fill!(::NDTensors.NoData, ::NDTensors.EmptyNumber)
```
Adding "Base.fill!(::NDTensors.NoData, ::Any) = 0  " is enough to solve the issue.

mtfishman:
@LHerviou does this kind of definition work:
```julia
Base.zero(T::EmptyTensor) = T
```
Seems like we may as well preserve that it is `Empty`. If so I can add that to NDTensors.jl.

LHerviou:
It seems to work perfectly fine.
A minor comment: would it not be safer to return a copy/similar of T? To be sure that the original tensor is not incremented.

mtfishman:
Good point, `copy(T)` makes sense.

mtfishman:
Actually, `EmptyTensor` is immutable so a `copy` isn't necessary.

LHerviou:
Nevermind me then.
While I am at the minor "minibug" due to EmptyTensors, dropzeros crash for
EmptyTensors due to eachnzblock/offsets not being defined.
Something like

(to put in
https://github.com/ITensor/ITensors.jl/blob/main/NDTensors/src/tensorstorage/tensorstorage.jl
)
"""
function ITensors.NDTensors.blockoffsets(::EmptyStorage{T,
NDTensors.BlockSparse{Q, Vector{R}, N}}) where {T, Q, R, N}
  return ITensors.NDTensors.BlockOffsets{N}()
end
"""
fixes that minor bug.

mtfishman:
I could believe that, that seems reasonable to add.

src/ITensors.jl
```diff
@@ -182,3 +182,17 @@ end
...
+#
+# function ITensors.NDTensors.datatype(
+#   ::ITensors.EmptyStorage{Float64,ITensors.NDTensors.BlockSparse{Float64,Vector{Float64},N}}
+# ) where {N}
+#   return Vector{Float64}
+# end
+#
+# function ITensors.NDTensors.datatype(T::ITensors.NDTensors.TensorStorage{Float64})
+#   return typeof(data(T))
+# end
```
mtfishman, line 196:
```suggestion
```

src/ITensors.jl
```diff
@@ -182,3 +182,17 @@ end
...
+
+Base.fill!(::NDTensors.NoData, ::Any) = 0
```
mtfishman, line 198: This is definition seems a bit funny, is this needed?

src/infinitempomatrix.jl
```diff
@@ -44,111 +44,191 @@ function InfiniteMPOMatrix(data::Vector{Matrix{ITensor}}, translator::Function)
...
-    Hm = Hms[n]
-    inds_n = TupleT[]
-    lx, ly = size(Hm)
-    @assert lx == ly
-    for iy in (ly - 1):-1:1
-      ix = iy + 1
-      il = ix < lx ? commonind(Hm[ix, iy], Hms[n - 1][ix + 1, iy + 1]) : dag(il0)
-      ir = iy > 1 ? commonind(Hm[ix, iy], Hms[n + 1][ix - 1, iy - 1]) : il0
-      push!(inds_n, (il, ir))
+function convert_itensor_33matrix(tensor; leftdir=ITensors.In, kwargs...)
```
mtfishman, line 151: Could you explain what these functions are doing?

src/infinitempomatrix.jl
```diff
@@ -44,111 +44,191 @@ function InfiniteMPOMatrix(data::Vector{Matrix{ITensor}}, translator::Function)
...
-  #
-  #  Create new tags with proper cell and link numbers.  Also daisy chain
-  #  all the indices so right index at j = dag(left index at j+1)
-  #
-  for j in 1:nsites(Hm)
-    newTag = "Link,c=$(getcell(site_inds[j])),l=$(getsite(site_inds[j]))"
-    ir = replacetags(irs[j], tags(irs[j]), newTag) #new right index
-    Hi[j] = replaceinds(Hi[j], [irs[j]], [ir])
-    Hi[j + 1] = replaceinds(Hi[j + 1], [ils[j + 1]], [dag(ir)])
+function apply_tensor(A::Array{ITensor,N}, B::ITensor...) where {N}
```
mtfishman, line 215: This doesn't really seem like it needs to be a function. For example it could just be done with:
```julia
map(a -> contract(a, B...), A)
```
if I understand it correctly.

src/infinitempomatrix.jl
```diff
@@ -44,111 +44,191 @@ function InfiniteMPOMatrix(data::Vector{Matrix{ITensor}}, translator::Function)
...
+#   A::T1, B::T2
+# ) where {T1<:ITensors.NDTensors.EmptyTensor,T2<:ITensors.NDTensors.EmptyTensor}
+#   ndims(A) != ndims(B) && throw(
+#     ITensors.DimensionMismatch("cannot add ITensors with different numbers of indices")
+#   )
+#   indices = inds(A)
+#   length(commoninds(indices, B)) != length(inds(B)) &&
+#     error("cannot add ITensors with different indices")
+#   return ITensor(promote_type(eltype(A), eltype(B)), indices)
+# end
```
mtfishman, line 234:
```suggestion
```

src/models/models.jl
```diff
@@ -110,22 +116,112 @@ function InfiniteMPOMatrix(model::Model, s::CelledVector, translator::Function;
...
   end
-  #return mpos
-  return InfiniteMPOMatrix(mpos, translator)
+  #unify_indices and add virtual indices to the empty tensors
+  mpos = InfiniteMPOMatrix(mpos, translator)
+  for x in 1:N
+    sp = prime(s[x])
+    sd = dag(s[x])
+    left_inds = [
+      only(uniqueinds(mpos.data.data[x][j, 1], mpos.data.data[x][1, 1])) for
```
mtfishman, line 169: Reaching into the data structures like this is definitely not recommended/encouraged, is there a better way to do this?

src/vumps_mpo.jl
```diff
@@ -1,3 +1,9 @@
+function ITensors.NDTensors.contraction_output(
+  A::NDTensors.EmptyTensor, B::NDTensors.DiagBlockSparseTensor, label
+)
+  return ITensor(eltype(B), label)
+end
+
```
mtfishman, line 6: I think there would be a better way to achieve this (at the very least, the `ITensor` type shouldn't be used in `NDTensors` code).

src/infinitempomatrix.jl
mtfishman: Or `contract.(A, Ref.(B)...)` should also work.

mtfishman:
@LHerviou sorry for the slow response on this. I added a definition for `blockoffsets` for tensors with `EmptyStorage` in https://github.com/ITensor/ITensors.jl/pull/1172, is that the last thing that was needed on the ITensor side for this PR?

src/ITensors.jl
mtfishman: I forgot this was fixing the issue with `zero` of an `EmptyStorage`.

It looks like this is sufficient:
```julia
fill!(::NoData, ::EmptyNumber) = NoData()
```
and is a bit safer/more specific. I'll add that to NDTensors.

mtfishman:
Oops I forgot about the issue with `zero` of an `ITensor` with `EmptyStorage`:
```julia
julia> zero(ITensor(Index(2)))
ERROR: MethodError: no method matching fill!(::NDTensors.NoData, ::NDTensors.EmptyNumber)

Closest candidates are:
  fill!(::Union{LinearAlgebra.Hermitian{T, S}, LinearAlgebra.Symmetric{T, S}} where {T, S}, ::Any)
   @ LinearAlgebra ~/.julia/juliaup/julia-1.9.2+0.aarch64.apple.darwin14/share/julia/stdlib/v1.9/LinearAlgebra/src/symmetric.jl:305
  fill!(::Union{LinearAlgebra.Bidiagonal, LinearAlgebra.Diagonal, LinearAlgebra.SymTridiagonal, LinearAlgebra.Tridiagonal}, ::Any)
   @ LinearAlgebra ~/.julia/juliaup/julia-1.9.2+0.aarch64.apple.darwin14/share/julia/stdlib/v1.9/LinearAlgebra/src/special.jl:358
  fill!(::Union{SparseArrays.AbstractCompressedVector, SparseArrays.AbstractSparseMatrixCSC}, ::Any)
   @ SparseArrays ~/.julia/juliaup/julia-1.9.2+0.aarch64.apple.darwin14/share/julia/stdlib/v1.9/SparseArrays/src/sparsevector.jl:2258
  ...

Stacktrace:
 [1] fill!(S::NDTensors.EmptyStorage{NDTensors.EmptyNumber, NDTensors.Dense{NDTensors.EmptyNumber, Vector{NDTensors.EmptyNumber}}}, v::NDTensors.EmptyNumber)
   @ NDTensors ~/.julia/dev/ITensors/NDTensors/src/tensorstorage/tensorstorage.jl:84
 [2] fill!(T::NDTensors.EmptyTensor{NDTensors.EmptyNumber, 1, Tuple{Index{Int64}}, NDTensors.EmptyStorage{NDTensors.EmptyNumber, NDTensors.Dense{NDTensors.EmptyNumber, Vector{NDTensors.EmptyNumber}}}}, α::NDTensors.EmptyNumber)
   @ NDTensors ~/.julia/dev/ITensors/NDTensors/src/tensor/tensor.jl:238
 [3] zero(x::NDTensors.EmptyTensor{NDTensors.EmptyNumber, 1, Tuple{Index{Int64}}, NDTensors.EmptyStorage{NDTensors.EmptyNumber, NDTensors.Dense{NDTensors.EmptyNumber, Vector{NDTensors.EmptyNumber}}}})
   @ Base ./abstractarray.jl:1209
 [4] zero(T::ITensor)
   @ ITensors ~/.julia/packages/ITensors/wm7BG/src/itensor.jl:680
 [5] top-level scope
   @ REPL[2]:1
```
As I showed above, this can be fixed with:
```julia
fill!(::NoData, ::EmptyNumber) = NoData()
```
I'll add that to NDTensors.

src/ITensors.jl
LHerviou: Removed

src/ITensors.jl
LHerviou: I had to keep that in (with the NoData()), it seems the newest version of ITensors does not yet have it?

src/infinitempomatrix.jl
LHerviou: Added docstrings for these functions.
They are just converting the tensors of a MPO/iMPO into 3x3 (or 1x3 or 3x1) matrices of MPOs

src/infinitempomatrix.jl
LHerviou: Removed the function and replaced it in the test.
It was convenient though not necessary indeed.

src/infinitempomatrix.jl
LHerviou: Removed

src/models/models.jl
LHerviou: So two things here
- I noticed a significant slow down/cost to do the operations at the mpos level, weirdly. Again the translatecell calls seem to have a small effect on performances (mainly because we are calling this _a_lot_, Normally the new functions should be better, though it is still visible.
I nonetheless removed all the direct accesses that are not necessary (see below)
- For some operations, it is required as we are modifying the tensors. Modifying the tensors at the global MPO level did not seem to work.

src/vumps_mpo.jl
LHerviou: Indeed, I replaced the output by NDTensors.EmptyTensor(eltype(B), label) which seems to work just as well.

I also moved it to ITensors.jl as it should be removed at some point.

LHerviou:
Hej @mtfishman

No worries for the delays, and I apologize also, I was in a conference. Quite a few people expressed interest in the compression schemes and other similar things.

In any case, I answered all comments. I am trying to see whether we can limit efficiently the number of calls to translatecell

src/infinitempomatrix.jl
```diff
@@ -44,111 +44,212 @@ function InfiniteMPOMatrix(data::Vector{Matrix{ITensor}}, translator::Function)
...
+end
+
+function ITensors.siteinds(A::InfiniteMPOMatrix)
+  return CelledVector(
+    [dag(only(filterinds(A[x][1, 1]; plev=0, tags="Site"))) for x in 1:nsites(A)],
+    translator(A),
+  )
+end
+
+function ITensors.splitblocks(H::InfiniteMPOMatrix)
```
mtfishman, line 62: Needs a `H = copy(H)` so it doesn't modify the input.

src/ITensors.jl
```diff
@@ -182,3 +182,13 @@ end
 # TODO: make this definition AbstractMPS
 # Handle orthogonality center correctly
 Base.getindex(ψ::MPS, r::UnitRange{Int}) = MPS([ψ[n] for n in r])
+
+#TODO Remove if everything is working nicely
+#Was still crashing on my laptop after updating ITensors
+Base.fill!(::NDTensors.NoData, ::Any) = NDTensors.NoData()
```
mtfishman, line 188: I added a similar definition in https://github.com/ITensor/ITensors.jl/pull/1173, is the definition there not sufficient?

src/ITensors.jl
```diff
@@ -182,3 +182,13 @@ end
...
+
+function ITensors.NDTensors.contraction_output(
+  A::NDTensors.EmptyTensor, B::NDTensors.DiagBlockSparseTensor, label
+)
+  return NDTensors.EmptyTensor(eltype(B), label)
```
mtfishman, line 193: I think `promote_type(eltype(A), eltype(B))` instead of `eltype(B)` would be more general and robust.

I would argue for moving this to NDTensors.jl, but we are working on removing the `EmptStorage` type in favor of a much simpler design in https://github.com/ITensor/ITensors.jl/pull/1161.

src/infinitempomatrix.jl
```diff
@@ -44,111 +44,212 @@ function InfiniteMPOMatrix(data::Vector{Matrix{ITensor}}, translator::Function)
...
+  Output: the 3x3 matrix of tensors
+  Optional arguments: leftdir: the direction of the left_index of the matrix to select it properly. Default to Itensors.In
+                      leftindex: instead one can directly provide the left index. Default to nothing
+                      siteindex: instead of relying on the tag, one can provide the two site indices. Default to tag filtering.
+                      left_tags: if we want to change the tags of the left indices. Default to tags(leftindex).
+                      right_tags: if we want to change the tags of the right indices. Default to tags(rightindex).
+"""
+function convert_itensor_33matrix(tensor; kwargs...)
+  @assert order(tensor) == 4
+  leftind = get(kwargs, :leftindex, nothing)
```
mtfishman, line 179: We are generally moving away from the style of using `get` to access keyword arguments, instead you should define the function like:
```julia
function convert_itensor_33matrix(tensor; leftindex=nothing, ...)
```

src/infinitempomatrix.jl
```diff
@@ -44,111 +44,212 @@ function InfiniteMPOMatrix(data::Vector{Matrix{ITensor}}, translator::Function)
...
+    error(
+      "Conversion of ITensor into matrix of ITensor not planned for this type of tensors"
     )
   end
-  IN = Hm[N + 1, N + 1] * onehot(T, dag(il0) => 1) * onehot(T, il => dim(il)) #I op in the bottom left of Hm
-  H, ir = directsum(H => ir, IN => il0; tags="Link,right") #1-D directsum to the put I in the bottom row, to the right of M[1]
+end
+
+"""
+    build_three_projectors_from_index(is::Index; kwargs...)
```
mtfishman, line 118: What about `local_mpo_block_projectors`?

mtfishman: To indicate what the projectors are being used for, i.e. to transform an MPO tensor into a local MPO block structure.

src/infinitempomatrix.jl
```diff
@@ -44,111 +44,212 @@ function InfiniteMPOMatrix(data::Vector{Matrix{ITensor}}, translator::Function)
...
+  such that we can then easily compress it. Note that for most of our tensors, the upper triangular part will be 0.
+  Input: tensor the four leg tensors
+  Output: the 3x3 matrix of tensors
+  Optional arguments: leftdir: the direction of the left_index of the matrix to select it properly. Default to Itensors.In
+                      leftindex: instead one can directly provide the left index. Default to nothing
+                      siteindex: instead of relying on the tag, one can provide the two site indices. Default to tag filtering.
+                      left_tags: if we want to change the tags of the left indices. Default to tags(leftindex).
+                      right_tags: if we want to change the tags of the right indices. Default to tags(rightindex).
+"""
+function convert_itensor_33matrix(tensor; kwargs...)
```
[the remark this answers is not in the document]
mtfishman: Also, what about calling it `local_mpo_blocks`?

src/infinitempomatrix.jl
```diff
@@ -44,111 +44,212 @@ function InfiniteMPOMatrix(data::Vector{Matrix{ITensor}}, translator::Function)
...
+    elseif length(temp) == 1
+      append!(right_links, temp)
+    else
+      error("ITensor does not seem to be of the correct order")
+    end
+  end
+  return left_links, right_links
+end
+
+function convert_itensor_to_itensormatrix(tensor; kwargs...)
```
mtfishman, line 105: Same comment as below, rename to something more specific to the application, like `local_mpo_blocks`. If we pass indices, for example as a tuple, we could dispatch on the length of the tuple, for example:
```julia
function local_mpo_blocks(t::ITensor, inds::NTuple{4,Index})
  # Implementation for bulk
end

function local_mpo_blocks(t::ITensor, inds::NTuple{3,Index})
  # Implementation for boundaries
end
```

src/infinitempomatrix.jl
```diff
@@ -44,111 +44,212 @@ function InfiniteMPOMatrix(data::Vector{Matrix{ITensor}}, translator::Function)
...
+  return InfiniteMPOMatrix(data, m, n, false)
+end
+
+function InfiniteMPOMatrix(data::CelledVector{Matrix{ITensor}})
+  return InfiniteMPOMatrix(data, 0, size(data)[1], false)
+end
+
+function ITensors.siteinds(A::InfiniteMPOMatrix)
+  return CelledVector(
+    [dag(only(filterinds(A[x][1, 1]; plev=0, tags="Site"))) for x in 1:nsites(A)],
```
mtfishman, line 57: We should avoid using `filterinds` with `tags`, not all site indices will have `"Site"` tags.

test/test_iMPOConversions.jl
```diff
@@ -1,6 +1,35 @@
 using ITensors
 using ITensorInfiniteMPS
 using Test
+
+#With the new definition of InfiniteMPOMatrix, the MPO is better behaved, and hence we need to be a bit more careful
+function special_expect(ψ::InfiniteCanonicalMPS, h::InfiniteSum{MPO})
```
mtfishman, line 6: Do we need both `expect` and `special_expect`? It sounds like the reason for defining `special_expect` is that `InfiniteMPOMatrix` changed a bit, so then should we just remove the old version of `expect` and rename this to `expect`?

src/infinitempomatrix.jl
LHerviou: Ah, I did not know that. 
I was precisely forcing myself to use kwargs as I was seeing it in your code.
I will update these functions accordingly.

Do you have a document for these conventions?

JanReimers: Hi Loic, There is an excellent writeup on argument handling here: https://itensor.github.io/ITensors.jl/dev/DeveloperGuide.html

test/test_iMPOConversions.jl
mtfishman: I realized after I commented that this is just used in tests so the name doesn't matter so much, but I would use `_expect` instead of `special_expect` since "special" isn't very descriptive.

test/test_iMPOConversions.jl
```diff
@@ -32,6 +61,42 @@ function ITensors.expect(ψ::InfiniteCanonicalMPS, h::InfiniteMPO)
...
+  l = commoninds(ψ.AL[0], ψ.AL[1])
+  L = map(a -> contract(a, δ(l, dag(prime(l)))), L)
+  r = commoninds(ψ.AR[Ncell + 1], ψ.AR[Ncell])
+  R = map(a -> contract(a, δ(r, dag(prime(r)))), R)
+  L = map(a -> contract(a, ψ.C[0], dag(prime(ψ.C[0]))), L)
+  for j in 1:nsites(ψ)
+    temp = map(a -> contract(a, ψ.AR[j], dag(prime(ψ.AR[j]))), h[j])
+    L = L * temp
+  end
+  return (L * R)[1][1]#ITensorInfiniteMPS.scalar_product(L, R)[1]
```
mtfishman, line 97: Could you remove the trailing comment?

Also, it may be better to use `(L * R)[][]` since then it will check that they are actually scalar values as expected, for example:
```julia
julia> randn(1)[]
-0.5989126360206526

julia> randn(2)[]
ERROR: BoundsError: attempt to access 2-element Vector{Float64} at index []
Stacktrace:
 [1] throw_boundserror(A::Vector{Float64}, I::Tuple{})
   @ Base ./abstractarray.jl:744
 [2] checkbounds
   @ ./abstractarray.jl:709 [inlined]
 [3] _getindex
   @ ./abstractarray.jl:1328 [inlined]
 [4] getindex(::Vector{Float64})
   @ Base ./abstractarray.jl:1296
 [5] top-level scope
   @ REPL[6]:1

julia> randn(1)[1]
-1.1477415235068003

julia> randn(2)[1]
-0.5202834767819509
```

src/infinitempo.jl
```diff
@@ -13,3 +13,131 @@ end
...
+    )
+    middle = cb * middle
+    comb, comb_ind = directsum(
+      top => uniqueinds(top, dag(right_link)),
+      middle => uniqueinds(middle, dag(right_link)),
+      bottom => uniqueinds(bottom, dag(right_link));
+      tags=[tags(right_link)],
+    )
+    Hcl.data.data[j] = Hcl[j] * comb
+    Hcl.data.data[j + 1] = dag(comb) * Hcl[j + 1]
```
mtfishman, line 119: Is there an issue using `Hcl[j] = Hcl[j] * comb`, etc. here?

src/infinitempomatrix.jl
```diff
@@ -44,111 +44,197 @@ function InfiniteMPOMatrix(data::Vector{Matrix{ITensor}}, translator::Function)
...
-  ir = dag(il0) #setup recusion.
-  for i in (N - 1):-1:1 #2-D directsums to place new blocks below and to the right.
-    H, (il, ir) = directsum(
-      H => (il, ir), Ms[i] => inds_array[i]; tags=["Link,left", "Link,right"]
+function InfiniteMPOMatrix(data::CelledVector{Matrix{ITensor}}, m::Int64, n::Int64)
+  return InfiniteMPOMatrix(data, m, n, false)
+end
+
+function InfiniteMPOMatrix(data::CelledVector{Matrix{ITensor}})
+  return InfiniteMPOMatrix(data, 0, size(data)[1], false)
```
mtfishman, line 52: Can use `size(data, 1)` instead of `size(data)[1]`.

src/infinitempomatrix.jl
```diff
@@ -44,111 +44,197 @@ function InfiniteMPOMatrix(data::Vector{Matrix{ITensor}}, translator::Function)
...
+#   if order(tensor) == 3
+#     return convert_itensor_3vector(tensor; kwargs...)
+#   elseif order(tensor) == 4
+#     return convert_itensor_33matrix(tensor; kwargs...)
+#   else
+#     error(
+#       "Conversion of ITensor into matrix of ITensor not planned for this type of tensors"
+#     )
+#   end
+# end
```
mtfishman, line 128:
```suggestion
```

mtfishman:
@LHerviou what do you think of changing the name of the type from `InfiniteMPOMatrix` to `InfiniteBlockMPO`? I think that is a bit more descriptive, and captures that the MPO is composed of [block matrices](https://en.wikipedia.org/wiki/Block_matrix).

LHerviou:
> @LHerviou what do you think of changing the name of the type from `InfiniteMPOMatrix` to `InfiniteBlockMPO`? I think that is a bit more descriptive, and captures that the MPO is composed of [block matrices](https://en.wikipedia.org/wiki/Block_matrix).

I do not mind that change, as long as I don't have to change it again after that.

LHerviou:
I think InfiniteBlockMPO is a good name. It is a MPO by block in the same sense block matrices exist.
So I have no objections.

I have no idea how many people (beyond me and some of my collaborators) used the InfiniteMPOMatrix anyway.

src/infiniteblockmpo.jl
```diff
@@ -226,3 +225,83 @@ function local_mpo_blocks(
...
+Fuse the non-site legs of the infiniteBlockMPO Hcl and the corresponding left L and right R environments.
+Preserve the corner structure.
+Essentially the inverse of splitblocks. It becomes useful for the very dense MPOs once get after compression sometimes.
+Hcl is modified on site, L and R are not.
+Input: Hcl the infinite MPO
+Optional  L   the left environment (a vector of  ITensors)
+R   the right environment (a vector of  ITensors)
+Output: Hcl, newL, newR the updated MPO and environments
+"""
+function combineblocks_linkinds(H::InfiniteBlockMPO; L=nothing, R=nothing)
```
mtfishman, line 241: Were `L` and `R` supposed to be keyword arguments? I'm ok either way, just checking, since they are positional arguments in the `InfiniteMPO` version of `combineblocks_linkinds`. Also that version doesn't provide default values, seems like we should try to keep the interfaces consistent between the two.

mtfishman: If we do decide they should be keyword arguments, we should choose better names for them, like `left_environment` and `right_environment`.

mtfishman:
@LHerviou this looks good to me now. Do you think this is ready to merge?

LHerviou:
Hi

I believe so.

LHerviou:
For the future: I guess I am letting @JanReimers finally do his own updates, then I will add all the iDMRG and so on using his version of compression.

mtfishman:
Thanks again!

LHerviou:
I will have a look either tonight or tomorrow.
The errors seem to imply something is missing in ITensors.NDTensors, but it might be something else

mtfishman:
Ah we have been changing some internals of `NDTensors`, and I guess this PR relied on some of those internals, so I could imagine there might be an issue there.

We are in the process of refactoring and simplifying `NDTensors` so hopefully it will mean some of the `NDTensors` overloads made in this PR can just be removed in the near future.

LHerviou:
@mtfishman  Ok, I found the source of the bug.

l269 and l272 of [[tensor_operations](https://github.com/ITensor/ITensors.jl/tree/main/src/tensor_operations)

/tensor_algebra.jl](https://github.com/ITensor/ITensors.jl/blob/main/src/tensor_operations/tensor_algebra.jl), 
you changed the way you define the directsum_projectors! from 
```
 D1[...] = true
``` 
to
"""
D1[...] = one(eltype(D1))
"""

In the end, the problem is that one(EmptyNumber) is not defined. I am testing some fixes right now.

mtfishman:
Ok interesting, thanks for looking into it. Does that mean an `EmptyStorage` is getting direct-summed somewhere?

LHerviou:
Yes, when I am fusing some indices.

LHerviou:
For info:
NDTensors.convert(::Type{NDTensors.EmptyNumber}, x::Float64) = NDTensors.EmptyNumber()
fixes the problem. Should I update ITensorInfiniteMPS or do you want to wait to fix NDTensors.

mtfishman:
I'm not sure that is something we want to define, since conversion shouldn't lose information (i.e. `Float64 -> ComplexF64` conversion is fine since it widens the type, but `ComplexF64 -> Float64` is only ok if the imaginary part is zero).

I wonder if `one(::EmptyNumber) = true` would fix it? I think that makes some sense as a definition anyway, since `EmptyNumber` is kind of a universal zero value and `true` is the smallest one value in Julia.

mtfishman:
We are replacing the `EmptyStorage` type with a better and simpler design in https://github.com/ITensor/ITensors.jl/pull/1161 so hopefully things like this will be easier to deal with after that is merged.

PR #79, opened 2023-10-27
Bump ITensors compat to fix `directsum` issue

mtfishman:
This bumps the compat of ITensors to v0.3.48, which includes a fix to `directsum` made in https://github.com/ITensor/ITensors.jl/pull/1221. This fixes an issue that was being discussed in #77 where some functionality that made use of `directum` for converting between MPO formats (`InfiniteBlockMPO` and `InfiniteMPO`) was broken.


mtfishman:
@LHerviou looks like tests are passing now.

PR #80, opened 2023-12-06
Add entanglement entropy examples

ryanlevy:
Added two functions to compare von Neumann entanglement entropy for the finite and infinite DMRG examples. 
See formula 21 at [arxiv:1810.07006](https://arxiv.org/pdf/1810.07006.pdf)

Also modified some tests to work for the latest ITensors dmrg api

### DIFF
--- a/examples/vumps/src/entropy.jl
+++ b/examples/vumps/src/entropy.jl
@@ -1,0 +1,32 @@
+using ITensors
+using ITensorInfiniteMPS
+
+# calculate -\sum_i rho_i log(rho_i)
+function calc_vN_entropy(S)
+  SvN, norm = 0.0, 0.0
+  for n in 1:dim(S, 1)
+    p = S[n, n]^2
+    SvN -= p * log(p)
+    norm += p
+  end
+  return SvN, norm
+end
+
+# calculate von Neumann entropy of a MPS cut between sites b-1 and b
+function entropy(ψ::MPS, b)
+  ψ = orthogonalize(ψ, b)
+  U, S, V = svd(ψ[b], (linkind(ψ, b - 1), siteind(ψ, b)))
+  SvN, norm = calc_vN_entropy(S)
+  @assert norm ≈ 1.0
+  return SvN
+end
+
+# calculate von Neumann entropy of an infinite MPS at cut b of the unit cell
+function entropy(ψ::InfiniteCanonicalMPS, b)
+  #calculate entropy
+  C = ψ.C[b]
+  Ũ, S, Ṽ = svd(C, inds(C)[1])
+  SvN, norm = calc_vN_entropy(S)
+  @assert norm ≈ 1.0
+  return SvN
+end
--- a/examples/vumps/src/entropy.jl
+++ b/examples/vumps/src/entropy.jl
@@ -2,7 +2,8 @@ using ITensors
 using ITensorInfiniteMPS
 
 # calculate -\sum_i rho_i log(rho_i)
-function calc_vN_entropy(S)
+function entropy(S::ITensor)
+  @assert order(S) == 2
   SvN, norm = 0.0, 0.0
   for n in 1:dim(S, 1)
     p = S[n, n]^2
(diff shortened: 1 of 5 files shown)


examples/vumps/vumps_hubbard_extended.jl
```diff
@@ -7,6 +7,31 @@ include(
...
 
+function entropy_finite(ψ_, b)
+  ψ = orthogonalize(ψ_, b)
+  U, S, V = svd(ψ[b], (linkind(ψ, b - 1), siteind(ψ, b)))
+  SvN = 0.0
+  for n in 1:dim(S, 1)
+    p = S[n, n]^2
+    SvN -= p * log(p)
+  end
+  return SvN
```
mtfishman, line 18: Seems like this part could be factored out into a separate function.

ryanlevy:
Thanks @mtfishman for the review, let me know what you think now

examples/vumps/vumps_hubbard_extended.jl
```diff
@@ -7,6 +7,32 @@ include(
...
+end
+
+function entropy_infinite(ψ, b)
+  #calculate entropy
+  C = ψ.C[b]
+  Ũ, S, Ṽ = svd(C, inds(C)[1])
+  SvN, norm = calc_vN_entropy(S)
+  @assert norm ≈ 1.0
+  return SvN
+end
```
mtfishman, line 35: Seems like you could use multiple dispatch here, i.e.:
```julia
entropy(ψ::MPS, b) = ...
entropy(ψ::InfiniteCanonicalMPS, b) = ...
```

examples/vumps/vumps_ising.jl
```diff
@@ -7,6 +7,33 @@ include(
...
+end
+
+function entropy_infinite(ψ, b)
+  #calculate entropy
+  C = ψ.C[b]
+  Ũ, S, Ṽ = svd(C, inds(C)[1])
+  SvN, norm = calc_vN_entropy(S)
+  @assert norm ≈ 1.0
+  return SvN
+end
```
mtfishman, line 35: We shouldn't define these in multiple files. Better to make a common file with one set of definitions, maybe in `ITensorInfiniteMPS.jl/examples/vumps/src/entropy.jl`. Repeating the same code multiple times adds maintenance burden.

mtfishman: Of course it would be better to have common sets of entropy definitions somewhere in the ITensor library which could be used here but that is a larger issue, not ready to commit to that yet before thinking it through.

examples/vumps/src/entropy.jl
```diff
@@ -0,0 +1,32 @@
+using ITensors
+using ITensorInfiniteMPS
+
+# calculate -\sum_i rho_i log(rho_i)
+function calc_vN_entropy(S)
```
mtfishman, line 5: Could just define this as `entropy(S::ITensor)`.

examples/vumps/vumps_hubbard_extended.jl
```diff
@@ -1,11 +1,13 @@
...
-    pkgdir(ITensorInfiniteMPS), "examples", "vumps", "src", "vumps_subspace_expansion.jl"
-  ),
-)
+base_path = joinpath(pkgdir(ITensorInfiniteMPS), "examples", "vumps", "src")
+src_files = [
+  joinpath(base_path, "vumps_subspace_expansion.jl"), joinpath(base_path, "entropy.jl")
+]
+for f in src_files
+  include(f)
+end
```
mtfishman, line 10:
```suggestion
src_files = [
  "vumps_subspace_expansion.jl",
  "entropy.jl",
]
for f in src_files
  include(joinpath(base_path, f))
end
```

ryanlevy: I prefer the original just because it allows a user to add their own files to include in `src_files` without having to deal with `base_path`. If you still prefer this cleaner suggested version I'm happy to change it though

mtfishman: I would rather use the simpler version, users can edit it as they see fit.

mtfishman:
Thanks!

PR #88, opened 2024-10-11
Fix 2d heisenberg example and add warning

ryanlevy:
As reported in https://itensor.discourse.group/t/vumps-example-2d-heisenberg-model-interaction-term/2052?u=ryanlevy

The 2D heisenberg example misses some operators, which is fixed in this version. I've also included a warning to help users catch this in the future, happy to take suggestions on alternative "fixes"

### DIFF
--- a/examples/vumps/vumps_2d_heisenberg.jl
+++ b/examples/vumps/vumps_2d_heisenberg.jl
@@ -34,9 +34,9 @@ function ITensorInfiniteMPS.unit_cell_terms(::Model"heisenberg2D"; width)
   opsum = OpSum()
   for i in 1:width
     # Vertical
-    opsum += -0.5, "S+", i, "S-", mod(i, width) + 1
-    opsum += -0.5, "S-", i, "S+", mod(i, width) + 1
-    opsum += "Sz", i, "Sz", mod(i, width) + 1
+    opsum += -0.5, "S+", i, "S-", i + 1
+    opsum += -0.5, "S-", i, "S+", i + 1
+    opsum += "Sz", i, "Sz", i + 1
     # Horizontal
     opsum += -0.5, "S+", i, "S-", i + width
     opsum += -0.5, "S-", i, "S+", i + width
--- a/examples/vumps/vumps_2d_heisenberg.jl
+++ b/examples/vumps/vumps_2d_heisenberg.jl
@@ -35,17 +35,17 @@ function ITensorInfiniteMPS.unit_cell_terms(::Model"heisenberg2D"; width, yperio
   opsum = OpSum()
   for i in 1:width
     # Vertical
-    opsum += -0.5, "S+", i, "S-", i + 1
-    opsum += -0.5, "S-", i, "S+", i + 1
+    opsum -= 0.5, "S+", i, "S-", i + 1
+    opsum -= 0.5, "S-", i, "S+", i + 1
     opsum += "Sz", i, "Sz", i + 1
     # Horizontal
-    opsum += -0.5, "S+", i, "S-", i + width
-    opsum += -0.5, "S-", i, "S+", i + width
+    opsum -= 0.5, "S+", i, "S-", i + width
+    opsum -= 0.5, "S-", i, "S+", i + width
     opsum += "Sz", i, "Sz", i + width
   end
   if yperiodic
-    opsum += -0.5, "S+", 1, "S-", width
-    opsum += -0.5, "S-", 1, "S+", width
+    opsum -= 0.5, "S+", 1, "S-", width
+    opsum -= 0.5, "S-", 1, "S+", width
     opsum += "Sz", 1, "Sz", width
   end
   return opsum
--- a/examples/vumps/vumps_2d_heisenberg.jl
+++ b/examples/vumps/vumps_2d_heisenberg.jl
@@ -34,9 +34,9 @@ function ITensorInfiniteMPS.unit_cell_terms(::Model"heisenberg2D"; width)
   opsum = OpSum()
   for i in 1:width
     # Vertical
-    opsum += -0.5, "S+", i, "S-", mod(i + 1, width)
-    opsum += -0.5, "S-", i, "S+", mod(i + 1, width)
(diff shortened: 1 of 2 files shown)


src/models/models.jl
```diff
@@ -245,6 +245,12 @@ function infinite_terms(opsum::OpSum; kwargs...)
   # stores all terms starting on site `i`.
   opsum_cell_dict = groupreduce(minimum ∘ ITensors.sites, +, opsum)
   nsites = maximum(keys(opsum_cell_dict))
+  # check that we don't have terms we will ignore
+  dropped = filter(x -> x <= 0, keys(opsum_cell_dict))
```
mtfishman, line 249: I don't remember all of the details of this code, is this checking if there is a term that has a site that is less than 1? Maybe that should just be an error, it seems weird to have terms that aren't valid that we just ignore.

examples/vumps/vumps_2d_heisenberg.jl
```diff
@@ -34,9 +34,9 @@ function ITensorInfiniteMPS.unit_cell_terms(::Model"heisenberg2D"; width)
   opsum = OpSum()
   for i in 1:width
     # Vertical
-    opsum += -0.5, "S+", i, "S-", mod(i + 1, width)
-    opsum += -0.5, "S-", i, "S+", mod(i + 1, width)
-    opsum += "Sz", i, "Sz", mod(i + 1, width)
+    opsum += -0.5, "S+", i, "S-", mod(i, width) + 1
+    opsum += -0.5, "S-", i, "S+", mod(i, width) + 1
+    opsum += "Sz", i, "Sz", mod(i, width) + 1
```
mtfishman, line 39: Since `i` goes up to `width`, doesn't that mean `mod(i, width)` doesn't do anything?

src/models/models.jl
ryanlevy: A few lines down you have `opsum_cell = [opsum_cell_dict[j] for j in 1:nsites]` where it is hardcoded to do sites 1 to nsites. Happy to switch this to an error and just enforce a sensible unit cell definition

mtfishman: Yes an error would be better.

ryanlevy:
I'm unable to replicate the errors from the CI at the moment (on 1.10.4), all tests pass locally for me

mtfishman:
It looks like it may be specific to Julia 1.11, it was just released and the CI started testing against that automatically.

mtfishman:
Here is a minimal code that reproduces the test failure in Julia 1.11:
```julia
using ITensors: ITensor, Index
i = Index(2)
a = reshape([ITensor(), ITensor(i)], 1, 2)
b = reshape([ITensor(2.0), ITensor(2.0, i)], 2, 1)
a * b
```
which leads to an error:
```julia
ERROR: StackOverflowError:
Stacktrace:
      [1] promote_result(::Type, ::Type, ::Type{NDTensors.EmptyNumber}, ::Type{Bool})
        @ Base ./promotion.jl:338
      [2] promote_type
        @ ./promotion.jl:318 [inlined]--- the above 2 lines are repeated 79982 more times ---
```
so it is related to performing matrix multiplication of matrices of ITensors where some of the ITensors are not initialized.

mtfishman:
It looks like there were a few missing `promote_rule` definitions for `EmptyNumber`, defining the following:
```julia
using NDTensors: EmptyNumber
Base.promote_rule(::Type{T}, ::Type{EmptyNumber}) where {T<:Number} = T
Base.promote_rule(::Type{Bool}, ::Type{EmptyNumber}) = Bool
```
seems to fix the issue. I'll make a PR to `NDTensors.jl`.

mtfishman:
Here are more minimal examples that error with stack overflows without those new definitions:
```julia
promote_type(Bool, EmptyNumber)
promote_type(EmptyNumber, Bool)
```

ryanlevy:
Thanks @mtfishman ! I can check things in InfiniteMPS once that's sorted (there looks like some deprecated `Vararg` calls I can open a PR for later)

mtfishman:
https://github.com/ITensor/ITensors.jl/pull/1541 is now merged, that should fix the test failures.

examples/vumps/vumps_2d_heisenberg.jl
```diff
@@ -42,12 +43,17 @@ function ITensorInfiniteMPS.unit_cell_terms(::Model"heisenberg2D"; width)
     opsum += -0.5, "S-", i, "S+", i + width
     opsum += "Sz", i, "Sz", i + width
   end
+  if yperiodic
+    opsum += -0.5, "S+", 1, "S-", width
+    opsum += -0.5, "S-", 1, "S+", width
```
mtfishman, line 48:
```suggestion
    opsum -= 0.5, "S+", 1, "S-", width
    opsum -= 0.5, "S-", 1, "S+", width
```
I see other code in the package avoids `-=`, I think it wasn't working in ITensors.jl at the time when I first wrote the package.

examples/vumps/vumps_2d_heisenberg.jl
```diff
@@ -30,24 +31,29 @@ initstate(n) = isodd(n) ? "↑" : "↓"
...
-function ITensorInfiniteMPS.unit_cell_terms(::Model"heisenberg2D"; width)
+function ITensorInfiniteMPS.unit_cell_terms(::Model"heisenberg2D"; width, yperiodic)
   opsum = OpSum()
   for i in 1:width
     # Vertical
-    opsum += -0.5, "S+", i, "S-", mod(i + 1, width)
-    opsum += -0.5, "S-", i, "S+", mod(i + 1, width)
-    opsum += "Sz", i, "Sz", mod(i + 1, width)
+    opsum += -0.5, "S+", i, "S-", i + 1
+    opsum += -0.5, "S-", i, "S+", i + 1
```
mtfishman, line 39:
```suggestion
    opsum -= 0.5, "S+", i, "S-", i + 1
    opsum -= 0.5, "S-", i, "S+", i + 1
```

mtfishman:
Looks good, thanks Ryan! I'll merge once tests pass.

PR #91, opened 2024-10-14
Minor cleanup of internals

ryanlevy:
This is the beginning of a small cleanup of some of the internals to match current upstream packages. 

Functions labeled `# added` are being included in a upstream PR and to be removed pending proper porting

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ITensorInfiniteMPS"
 uuid = "1dc1fb26-a137-4954-ae60-1bd4106e95ad"
 authors = ["Matthew Fishman <mfishman@flatironinstitute.org> and contributors"]
-version = "0.1.0"
+version = "0.2.0"
 
 [deps]
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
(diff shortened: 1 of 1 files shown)


mtfishman:
@ryanlevy sorry for the slow response on this, this slipped my mind. I started making a PR to fix #100 which turned out to be similar to this PR, what is needed for this PR to be completed? The changes you've made here look reasonable to me.

ryanlevy:
This is dependent on https://github.com/ITensor/ITensors.jl/pull/1544, then after that we can plan on a potential ITensorMPS.jl PR to fully disentangle the package (regarding slicing a MPS)

mtfishman:
Can we simplify things and first make a PR here that includes the changes that don't depend on https://github.com/ITensor/ITensors.jl/pull/1544?

mtfishman:
Look good to me, though mostly what I'm going by is that this PR removes code and the tests still pass, so hopefully the tests are comprehensive!

mtfishman:
Can you bump the version and register after merging?

ryanlevy:
> Look good to me, though mostly what I'm going by is that this PR removes code and the tests still pass, so hopefully the tests are comprehensive!

Seems so, the examples still work and get to the exact energies which is really what I'm basing it off of! 

> Can you bump the version and register after merging?

I've bumped to 0.2 and also increased the ITensors to 0.8

PR #95, opened 2024-10-30
Fix 2D Heisenberg Example

ryanlevy:
(Reported in the [discourse](https://itensor.discourse.group/t/vumps-example-2d-heisenberg-model-interaction-term/2052/9?u=ryanlevy)) In removing the `mod` of the previous PR the yperiodic term was incorrect - there are `width` Horizontal bonds and `width-1` vertical and 1 potential wrapping. This should be the correct terms (and looks nicer than the previous version)
With `yperiodic=true`
```
opsum = sum(
  -0.5 S+(1,) S-(2,)
  -0.5 S-(1,) S+(2,)
  1.0 Sz(1,) Sz(2,)
  -0.5 S+(1,) S-(5,)
  -0.5 S-(1,) S+(5,)
  1.0 Sz(1,) Sz(5,)
  -0.5 S+(2,) S-(3,)
  -0.5 S-(2,) S+(3,)
  1.0 Sz(2,) Sz(3,)
  -0.5 S+(2,) S-(6,)
  -0.5 S-(2,) S+(6,)
  1.0 Sz(2,) Sz(6,)
  -0.5 S+(3,) S-(4,)
  -0.5 S-(3,) S+(4,)
  1.0 Sz(3,) Sz(4,)
  -0.5 S+(3,) S-(7,)
  -0.5 S-(3,) S+(7,)
  1.0 Sz(3,) Sz(7,)
  -0.5 S+(4,) S-(1,)
  -0.5 S-(4,) S+(1,)
  1.0 Sz(4,) Sz(1,)
  -0.5 S+(4,) S-(8,)
  -0.5 S-(4,) S+(8,)
  1.0 Sz(4,) Sz(8,)
```

and `yperiodic=false`
```
opsum = sum(
  -0.5 S+(1,) S-(2,)
  -0.5 S-(1,) S+(2,)
  1.0 Sz(1,) Sz(2,)
  -0.5 S+(1,) S-(5,)
  -0.5 S-(1,) S+(5,)
  1.0 Sz(1,) Sz(5,)
  -0.5 S+(2,) S-(3,)
  -0.5 S-(2,) S+(3,)
  1.0 Sz(2,) Sz(3,)
  -0.5 S+(2,) S-(6,)
  -0.5 S-(2,) S+(6,)
  1.0 Sz(2,) Sz(6,)
  -0.5 S+(3,) S-(4,)
  -0.5 S-(3,) S+(4,)
  1.0 Sz(3,) Sz(4,)
  -0.5 S+(3,) S-(7,)
  -0.5 S-(3,) S+(7,)
  1.0 Sz(3,) Sz(7,)
  -0.5 S+(4,) S-(8,)
  -0.5 S-(4,) S+(8,)
  1.0 Sz(4,) Sz(8,)
)
```

### DIFF
--- a/examples/vumps/vumps_2d_heisenberg.jl
+++ b/examples/vumps/vumps_2d_heisenberg.jl
@@ -31,22 +31,6 @@ initstate(n) = isodd(n) ? "↑" : "↓"
 s = infsiteinds("S=1/2", N; conserve_qns, initstate)
 ψ = InfMPS(s, initstate)
 
-function ITensorInfiniteMPS.unit_cell_terms(::Model"heisenberg2D"; width, yperiodic)
-  opsum = OpSum()
-  for i in 1:width
-    # Vertical
-    if i < width || yperiodic
-      opsum -= 0.5, "S+", i, "S-", mod(i, width) + 1
-      opsum -= 0.5, "S-", i, "S+", mod(i, width) + 1
-      opsum += "Sz", i, "Sz", mod(i, width) + 1
-    end
-    # Horizontal
-    opsum -= 0.5, "S+", i, "S-", i + width
-    opsum -= 0.5, "S-", i, "S+", i + width
-    opsum += "Sz", i, "Sz", i + width
-  end
-  return opsum
-end
 model = Model("heisenberg2D")
 
 # Form the Hamiltonian
@@ -72,6 +56,11 @@ subspace_expansion_kwargs = (cutoff=cutoff, maxdim=maxdim)
 
 energy_infinite = expect(ψ, H)
 @show energy_infinite
+@show sum(energy_infinite) / width
 
+energy_approx_exact = reference(model, Observable("energy"); width, yperiodic)
+@show energy_approx_exact
+
+@show isapprox(sum(energy_infinite) / width, energy_approx_exact, atol=1e-4)
 ## using JLD2
 ## jldsave("infmps.jld2"; ψ)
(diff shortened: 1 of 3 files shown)


mtfishman:
Thanks Ryan, can you add a test that it is outputting the correct Hamiltonian?

ryanlevy:
Since this is an example, what sort of test were you thinking of? (As a side note, this has me thinking an ITensorMPS tool that converts a opsum into an adjacency graph matrix might be useful)

mtfishman:
> Since this is an example, what sort of test were you thinking of?

We could split off the Hamiltonian definition into a separate file and then include it into this example and also into a test.

> (As a side note, this has me thinking an ITensorMPS tool that converts a opsum into an adjacency graph matrix might be useful)

That could be useful, though we'd have to think carefully about what it should output (i.e. does it include the operators and weights as well?), how to make it general to handle terms that are beyond 2-site, etc. Maybe better as an external tool. I think the same kind of thing can be accomplished by outputting a dictionary from the set of sites to the corresponding term of the Hamiltonian, which is easy enough to write, say using https://github.com/JuliaData/SplitApplyCombine.jl.

ryanlevy:
I've added the model internally and added a reference. The width=4 with OBC seems to work, but PBC the symmetry breaking in the unit cell is bad and we don't quite reach the correct value.

mtfishman:
> I've added the model internally and added a reference. The width=4 with OBC seems to work, but PBC the symmetry breaking in the unit cell is bad and we don't quite reach the correct value.

I was picturing that we could just check that some subset of the terms are correct, i.e. check the `OpSum` itself.

mtfishman:
Looks good, thanks!

PR #101, opened 2025-02-27
Gauge improvements

ryanlevy:
This implements two improvements:

- `finite_mps` can now have its gauge center be on the first site instead of the last site
- `orthogonalize` ensures a real Hermitian transfer matrix eigenvector thanks to `schursolve` (shout out @lkdvos for the tip)

The first change is a small speed improvement for long `finite_mps` calls as `correlation_matrix` will immediately move the orthogonality center to the first site

```
julia> @btime corrR = correlation_matrix(finite_mps(ψ,1:24; ortho="right"),"Cdagup","Cup"; sites=2:24+1);
  57.864 ms (900184 allocations: 146.93 MiB)

julia> @btime corrL = correlation_matrix(finite_mps(ψ,1:24; ortho="left"),"Cdagup","Cup"; sites=2:24+1);
  47.981 ms (764616 allocations: 117.28 MiB)
```

The second change includes the additional "gauge fixing" of the output of vumps in the Hubbard example, which should probably always be done so that the state you want to measure has a more consistent `C` matrix.

### DIFF
--- a/src/infinitecanonicalmps.jl
+++ b/src/infinitecanonicalmps.jl
@@ -252,7 +252,7 @@ end
 function finite_mps(
   ψ::InfiniteCanonicalMPS,
   range::AbstractRange;
-  ortho_lims::UnitRange=last(range):last(range),
+  ortho_lims::AbstractUnitRange=last(range):last(range),
 )
   @assert isone(step(range))
   @assert first(ortho_lims) == last(ortho_lims) # TODO: variable ortho_lims
--- a/src/infinitecanonicalmps.jl
+++ b/src/infinitecanonicalmps.jl
@@ -249,26 +249,28 @@ function InfMPS(s::CelledVector, f::Function)
   return ψ = InfiniteCanonicalMPS(ψL, ψC, ψR)
 end
 
-function finite_mps(ψ::InfiniteCanonicalMPS, range::AbstractRange; ortho::String="right")
+function finite_mps(
+  ψ::InfiniteCanonicalMPS,
+  range::AbstractRange;
+  ortho_lims::UnitRange=last(range):last(range),
+)
   @assert isone(step(range))
+  @assert first(ortho_lims) == last(ortho_lims) # TODO: variable ortho_lims
+  @assert first(ortho_lims) ∈ range
+
   N = length(range)
-  if ortho == "right"
-    ψ_finite = ψ.AL[range]
-    ψ_finite[N] *= ψ.C[last(range)]
-    l0 = linkind(ψ.AL, first(range) - 1 => first(range))
-    lN = linkind(ψ.AR, last(range) => last(range) + 1)
-    center = N + 1
-  elseif ortho == "left"
-    ψ_finite = (ψ.AR[range])
-    # C gets applied to the left here
-    ψ_finite[1] *= (ψ.C[first(range) - 1])
-    l0 = linkind(ψ.AL, first(range) - 1 => first(range))
-    lN = linkind(ψ.AR, last(range) => last(range) + 1)
-    center = 2
-  else
-    error("Unknown ortho option $ortho. Please use left or right")
+  ψ_finite = Vector{ITensor}(undef, N)
+  for i in first(range):first(ortho_lims)
+    ψ_finite[i] = ψ.AL[i]
+  end
+  ψ_finite[first(ortho_lims)] *= ψ.C[first(ortho_lims)]
+
+  for i in (last(ortho_lims) + 1):last(range)
+    ψ_finite[i] = ψ.AR[i]
   end
 
+  l0 = linkind(ψ.AL, first(range) - 1 => first(range))
+  lN = linkind(ψ.AR, last(range) => last(range) + 1)
   l̃0 = sim(l0)
   l̃N = sim(lN)
   δl0 = δ(dag(l̃0), l0)
@@ -276,7 +278,7 @@ function finite_mps(ψ::InfiniteCanonicalMPS, range::AbstractRange; ortho::Strin
   ψ_finite[1] *= δl0
   ψ_finite[N] *= dag(δlN)
   ψ_finite = MPS(
-    [dag(δl0); [ψ_finiteᵢ for ψ_finiteᵢ in ψ_finite]; δlN]; ortho_lims=center:center
+    [dag(δl0); ψ_finite; δlN]; ortho_lims=(first(ortho_lims) + 1):(last(ortho_lims) + 1)
   )
   return ψ_finite
 end
--- a/src/infinitecanonicalmps.jl
+++ b/src/infinitecanonicalmps.jl
@@ -249,21 +249,35 @@ function InfMPS(s::CelledVector, f::Function)
   return ψ = InfiniteCanonicalMPS(ψL, ψC, ψR)
 end
 
-function finite_mps(ψ::InfiniteCanonicalMPS, range::AbstractRange)
+function finite_mps(ψ::InfiniteCanonicalMPS, range::AbstractRange; ortho::String="right")
   @assert isone(step(range))
   N = length(range)
-  ψ_finite = ψ.AL[range]
-  ψ_finite[N] *= ψ.C[last(range)]
-  l0 = linkind(ψ.AL, first(range) - 1 => first(range))
+  if ortho == "right"
+    ψ_finite = ψ.AL[range]
+    ψ_finite[N] *= ψ.C[last(range)]
+    l0 = linkind(ψ.AL, first(range) - 1 => first(range))
+    lN = linkind(ψ.AR, last(range) => last(range) + 1)
+    center = N + 1
+  elseif ortho == "left"
+    ψ_finite = (ψ.AR[range])
+    # C gets applied to the left here
+    ψ_finite[1] *= (ψ.C[first(range) - 1])
+    l0 = linkind(ψ.AL, first(range) - 1 => first(range))
+    lN = linkind(ψ.AR, last(range) => last(range) + 1)
+    center = 2
+  else
+    error("Unknown ortho option $ortho. Please use left or right")
+  end
+
   l̃0 = sim(l0)
-  lN = linkind(ψ.AR, last(range) => last(range) + 1)
   l̃N = sim(lN)
   δl0 = δ(dag(l̃0), l0)
   δlN = δ(dag(l̃N), lN)
   ψ_finite[1] *= δl0
   ψ_finite[N] *= dag(δlN)
-  ψ_finite = MPS([dag(δl0); [ψ_finiteᵢ for ψ_finiteᵢ in ψ_finite]; δlN])
-  set_ortho_lims!(ψ_finite, (N + 1):(N + 1))
+  ψ_finite = MPS(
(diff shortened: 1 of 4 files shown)


src/infinitecanonicalmps.jl
```diff
@@ -249,21 +249,35 @@ function InfMPS(s::CelledVector, f::Function)
   return ψ = InfiniteCanonicalMPS(ψL, ψC, ψR)
 end
 
-function finite_mps(ψ::InfiniteCanonicalMPS, range::AbstractRange)
+function finite_mps(ψ::InfiniteCanonicalMPS, range::AbstractRange; ortho::String="right")
```
mtfishman, line 252: It would make more sense to me to instead have a keyword argument `ortho_lims`, like the MPS constructor.

lkdvos, line 252: I'm guessing this is a bit of a convention thing, but setting options like this via `Symbol`s ensures that the compiler can see this. In particular, it can then typically constant propagate this entire function and eliminate the branching.
Since it's in a kwarg, you can always consider "standardizing" the input beforehand to allow both?
```suggestion
function finite_mps(ψ::InfiniteCanonicalMPS, range::AbstractRange; ortho=:right)
    _ortho = ortho isa Symbol ? ortho : Symbol(ortho)
```

(of course, probably does not matter all that much)

src/orthogonalize.jl
```diff
@@ -12,9 +13,15 @@ function right_orthogonalize(
...
   # all of the right environments using `eigsolve` and shifting unit cells
-  λ⃗₁ᴿᴺ, v⃗₁ᴿᴺ, eigsolve_info = eigsolve(T, v₁ᴿᴺ, 1, :LM; tol=tol)
+
+  # original eigsolve function, switch to schur which enforces real
+  #λ⃗₁ᴿᴺ, v⃗₁ᴿᴺ, eigsolve_info = eigsolve(T, v₁ᴿᴺ, 1, :LM; tol, eager=true)
+  TT, v⃗₁ᴿᴺ, λ⃗₁ᴿᴺ, eigsolve_info = schursolve(T, v₁ᴿᴺ, 1, :LM, Arnoldi(; tol))
   λ₁ᴿᴺ, v₁ᴿᴺ = λ⃗₁ᴿᴺ[1], v⃗₁ᴿᴺ[1]
 
+  if size(TT, 2) > 1 && TT[2, 1] != 0
+    @warn("Largest transfer matrix eigenvector is not real?")
```
mtfishman, line 23: Seems a bit funny to make this a question. Can't you check if it is real?

mtfishman: If it isn't a definitive check, it would be better to phrase it as `"Largest transfer matrix eigenvector might not be real."`.

src/orthogonalize.jl
```diff
@@ -12,9 +13,15 @@ function right_orthogonalize(
   # Start by getting the right eivenvector/eigenvalue of T
   # TODO: make a function `right_environments(::InfiniteMPS)` that computes
   # all of the right environments using `eigsolve` and shifting unit cells
-  λ⃗₁ᴿᴺ, v⃗₁ᴿᴺ, eigsolve_info = eigsolve(T, v₁ᴿᴺ, 1, :LM; tol=tol)
+
+  # original eigsolve function, switch to schur which enforces real
+  #λ⃗₁ᴿᴺ, v⃗₁ᴿᴺ, eigsolve_info = eigsolve(T, v₁ᴿᴺ, 1, :LM; tol, eager=true)
+  TT, v⃗₁ᴿᴺ, λ⃗₁ᴿᴺ, eigsolve_info = schursolve(T, v₁ᴿᴺ, 1, :LM, Arnoldi(; tol))
```
lkdvos, line 19:
```suggestion
  TT, v⃗₁ᴿᴺ, λ⃗₁ᴿᴺ, eigsolve_info = schursolve(T, v₁ᴿᴺ, 1, :LM, Arnoldi(; tol, eager=true))
```

src/orthogonalize.jl
ryanlevy: I guess it should be "Non-unique largest eigenvector" (or the eigenvector shouldn't be real)

lkdvos: I think what it really means is that it is not unique, this results in complex conjugate eigenvalue pairs (hence checking the `TT[2, 1]`, `TT[1:2,1:2] = [real(lambda) imag(lambda); -imag(lambda) real(lambda)]` or something like that).

src/infinitecanonicalmps.jl
ryanlevy: `ortho_lims` is a bit funny because they will be off by 1 due to the environment tensors. But we can just translate 1=>2 and N=> N+1

mtfishman: (But also as a convention we generally use strings for keyword arguments that select different options unless there is a performance reason, since it looks friendlier to users, but as you say we can just allow generic inputs and canonicalize the input.)

mtfishman:
Nice to see it get more general and simpler!

ryanlevy:
@lkdvos it seems that using `eager=true` causes the vector to be non-hermitian (at least higher than the Arnoldi tolerance)

lkdvos:
That's interesting... The only difference should be that it is checking every iteration if the tolerance has been reached, instead of waiting until the full krylov subspace is populated. Basically, it's probably not fair to expect that you reach the same tolerance for the eigenvector hermiticity as you would get for the eigenvector itself. If you don't put `eager=true`, you often just get a higher precision than you asked for, because you are doing additional iterations. However, in our cases the extra iterations are quite expensive, so this **strongly** affects performance

mtfishman:
Sounds like we should play around with different tolerances when checking for Hermiticity, I don't think there is anything magical about the current choice besides that it has worked so far for catching issues but not being too strict.

src/orthogonalize.jl
```diff
@@ -1,7 +1,13 @@
...
 # TODO: call as `orthogonalize(ψ, -∞)`
 # TODO: could use commontags(ψ) as a default for left_tags
 function right_orthogonalize(
-  ψ::InfiniteMPS; left_tags=ts"Left", right_tags=ts"Right", tol::Real=1e-12
+  ψ::InfiniteMPS;
+  left_tags=ts"Left",
+  right_tags=ts"Right",
+  tol::Real=1e-12,
+  eager=true,
+  htol::Real=1e-10,
```
mtfishman, line 10: I would call this `ishermitian_rtol`, no need for a short name here. Shouldn't we define it relative to `tol`, for example `ishermitian_rtol=10tol`?

Also if the Hermitian check is causing trouble, maybe we should just turn it into a warning.

mtfishman: Or better yet, I would call this `ishermitian_kwargs` and set a default like `ishermitian_kwargs=(; rtol=10tol)`.

ryanlevy:
I noticed a bug with `orthogonalize`, you can't call it twice because it adds tags which interfere with the next run
```julia
julia> p = orthogonalize(ψ.AL, : ; tol=1e-14)
       p2 = orthogonalize(p.AL, :)
ERROR: In `isapprox(::ITensor, ::ITensor)`, the indices of the ITensors do not
          match. The first ITensor has indices:

#....
```

src/orthogonalize.jl
ryanlevy: Sure thing, I'll do both

mtfishman:
> I noticed a bug with `orthogonalize`, you can't call it twice because it adds tags which interfere with the next run
> 
> ```julia
> julia> p = orthogonalize(ψ.AL, : ; tol=1e-14)
>        p2 = orthogonalize(p.AL, :)
> ERROR: In `isapprox(::ITensor, ::ITensor)`, the indices of the ITensors do not
>           match. The first ITensor has indices:
> 
> #....
> ```

Got it, I don't have time to look into that but feel free to make a PR.

src/orthogonalize.jl
```diff
@@ -1,7 +1,13 @@
...
 # TODO: call as `orthogonalize(ψ, -∞)`
 # TODO: could use commontags(ψ) as a default for left_tags
 function right_orthogonalize(
-  ψ::InfiniteMPS; left_tags=ts"Left", right_tags=ts"Right", tol::Real=1e-12
+  ψ::InfiniteMPS;
+  left_tags=ts"Left",
+  right_tags=ts"Right",
+  tol::Real=1e-12,
+  eager=true,
+  ishermitian_rtol::Real=tol * 100,
```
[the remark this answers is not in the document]
mtfishman: Then you can splat `ishermitian_kwargs` into the `ishermitian` call.

ryanlevy:
@mtfishman is this good to go?

src/orthogonalize.jl
```diff
@@ -24,11 +41,11 @@ function right_orthogonalize(
 
   # Fix the phase of the diagonal to make Hermitian
   v₁ᴿᴺ .*= conj(sign(v₁ᴿᴺ[1, 1]))
-  if !ishermitian(v₁ᴿᴺ; rtol=tol)
-    @show λ₁ᴿᴺ
-    @show v₁ᴿᴺ
+  if !ishermitian(v₁ᴿᴺ; ishermitian_kwargs...)
+    #@show λ₁ᴿᴺ
+    #@show v₁ᴿᴺ
```
mtfishman, line 46:
```suggestion
```

src/orthogonalize.jl
```diff
@@ -105,11 +122,16 @@ end
...
+  _, ψᴿ, _ = right_orthogonalize(ψ; left_tags=ts"", right_tags)
+  ψᴸ, C, λ = left_orthogonalize(ψᴿ; left_tags, right_tags)
   if λ ≉ one(λ)
     error("λ should be approximately 1 after orthogonalization, instead it is $λ")
   end
+  # tags have been added, remove them now
+  # This is broken until we fix the tags for C
+  #removetags!(ψᴸ, left_tags)
+  #removetags!(ψᴿ, right_tags)
+  #removetags!(C,addtags(right_tags,left_tags))
```
mtfishman, line 134: Can we just delete this for now?

src/infinitecanonicalmps.jl
```diff
@@ -249,21 +249,37 @@ function InfMPS(s::CelledVector, f::Function)
   return ψ = InfiniteCanonicalMPS(ψL, ψC, ψR)
 end
 
-function finite_mps(ψ::InfiniteCanonicalMPS, range::AbstractRange)
+function finite_mps(
+  ψ::InfiniteCanonicalMPS,
+  range::AbstractRange;
+  ortho_lims::UnitRange=last(range):last(range),
```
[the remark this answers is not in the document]
ryanlevy: `MPST` takes `UnitRange` and not `AbstractUnitRange` so we should keep it consistent right?

src/orthogonalize.jl
ryanlevy: Yeah my bad

src/infinitecanonicalmps.jl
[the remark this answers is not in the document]
mtfishman: I think for higher up functions it is better to allow more general inputs, if it fails lower down then that is fine.

mtfishman:
Can you bump the package version?

ryanlevy:
Sure thing!

